PR #36, opened 2024-12-18
chore: add logging; fixup some imports

gregory-j-baker:
Add some logging to authentication-strategy.ts


future-sir-frontend/app/utils/auth/authentication-strategy.ts
```diff
@@ -76,6 +77,7 @@ export abstract class BaseAuthenticationStrategy implements AuthenticationStrate
   protected readonly client: Client;
   protected readonly clientAuth: ClientAuth;
 
+  protected readonly log = LogFactory.getLogger(import.meta.url);
```
sebastien-comeau, line 80: I don't like it cause if you have multiple class in one file, they will use the same log category and you won't be able to easily distinguish the log when they are displayed.

PR #47, opened 2024-12-20
fix(frontend): return responses instead of throwing

gregory-j-baker:
Returning responses instead of throwing in loaders/actions. See <https://reactrouter.com/start/framework/navigating#redirect>.

### DIFF
--- a/frontend/app/routes/api/translations.ts
+++ b/frontend/app/routes/api/translations.ts
@@ -10,14 +10,14 @@ export async function loader({ request }: Route.LoaderArgs) {
   const buildRevision = url.searchParams.get('v');
 
   if (!lng || !ns) {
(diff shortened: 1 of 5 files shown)


sebastien-comeau (commented):
It's ok to throw responses, especially 404s.

[3. Throw data in loaders/actions](https://reactrouter.com/how-to/error-boundary#3-throw-data-in-loadersactions)

> There are exceptions to the rule in #&NoBreak;2, especially 404s.

[Throwing Responses](https://reactrouter.com/tutorials/address-book#throwing-responses)

sebastien-comeau (approved):
Returning `response` or r`edirect` make sense in `Resource Routes` since they are not meant to render HTML.

PR #53, opened 2024-12-24
feat(frontend): add menu component

Dario-Au:
Added a menu component that replaces the list of links on the public and protected index page

image
image


gregory-j-baker (approved):

sebastien-comeau:
I suggest that we use [Dropdown Menu](https://ui.shadcn.com/docs/components/dropdown-menu) since it's been used in CDCP. Shadcn/ui style [Radix UI Dropdown Menu](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) and provide the component source code that you can copy paste into Future SIR. You have to modify the style match our template once you copied the source code.

PR #69, opened 2025-01-10
refactor(frontend): app menu refactoring

sebastien-comeau:
### Description:

This PR introduces the following changes based on the diff:

1. **MenuItem Component:**
   - **Link Component Change**: The `InlineLink` has been replaced with `AppLink` in the `MenuItem` component. `AppLink` is preferred because it supports application i18n links and does not include any styling, making it more flexible for linking.
   - **Data-testid**: A `data-testid="menu-item"` attribute has been added to the `AppLink` for easier testing.

2. **Menu Component:**
   - **Dropdown Open State**: The dropdown open state is now managed by the Radix library, simplifying the handling of the menu’s open/closed state.
   - **DropdownMenuTrigger**: 
     - The `DropdownMenuTrigger` styling has been simplified, now directly using `cn` to manage the Tailwind CSS classes. The `focus:ring` classes are enhanced with additional focus-related styles, including `focus:ring-offset-2` and `aria-expanded` styles that update the button's appearance when the menu is open.
     - The button structure is unchanged, but the styling is updated to be more responsive and visually consistent with the open/close state of the menu.

3. **Chevron Icon:**
   - The `UpChevron` component has been removed.

These changes enhance the consistency of the dropdown component by streamlining styling, improving accessibility with the use of `aria-expanded`, and updating the link rendering mechanism to leverage `AppLink` for i18n support. Additionally, the Radix library now manages the dropdown open state for a more robust solution.

#### Links

- [ARIA States in Tailwind CSS](https://tailwindcss.com/docs/hover-focus-and-other-states#aria-states): Use the `aria-*` modifier to conditionally style things based on [ARIA attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes).


Dario-Au (changes requested):
Tests failed due to outdated snapshots for the Menu component

sebastien-comeau:
> Tests failed due to outdated snapshots for the Menu component

Right, I'm fixing them.

frontend/app/components/menu.tsx
```diff
@@ -1,25 +1,19 @@
...
   return (
-    <DropdownMenuItem className="text-md m-0 p-0 hover:!bg-slate-700">
-      <InlineLink
-        role="menuitem"
-        id="menu-item"
-        className="w-full px-3 py-2 text-white hover:bg-slate-300 hover:text-blue-950 focus:bg-slate-600 focus:text-white active:bg-slate-800 active:text-white"
-        {...props}
-      >
+    <DropdownMenuItem asChild className="cursor-pointer">
+      <AppLink data-testid="menu-item" {...props}>
```
gregory-j-baker, line 14: I'm not a fan of using `data-testid` in components like this. Instead, my preferred approach is to add the attribute in the test. Something like this:

```
const RoutesStub = createRoutesStub([
  {
    path: '/',
    Component: () => (
      <DropdownMenuItem data-testid="foo">
        ...
      </DropdownMenuItem>
    ),
  },
]);
```

Then run your assertions using the provided testid.

sebastien-comeau: Your approach will work since you can assign an attribute to it. However, when dealing with a component that has children composition and you need to perform assertions on its children, it's better to use data-testid.

gregory-j-baker: Use the testid to target the component being tested, then isolate any children as necessary using selectors.

Why unnecessarily clutter the production html if we don't have to?

sebastien-comeau: `data-testid` can be usefull when you're writing e2e tests. Another team could be responible to write those tests and could struggle to select elements. We had this issue with other projects before.

Dario-Au (approved):

PR #71, opened 2025-01-10
fix(frontend): vite dependency rollup issue

sebastien-comeau:
Use the WASM build [@rollup/wasm-node](https://www.npmjs.com/package/@rollup/wasm-node) instead of the native build with [Vite](https://vite.dev/).

> Rollup now includes native code that is automatically installed (and removed) as an [optional npm dependency](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) if your platform and architecture is supported. More precisely, Rollup has a list of optionalDependencies, each of which only install on a specific [os](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#os) and [cpu](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#cpu). If your system is not supported, you will receive an error message when starting Rollup that will tell you about your platform and architecture and gives you a list of supported ones. In that case, you can instead use @rollup/wasm-node as a platform-independent drop-in replacement.

https://rollupjs.org/migration/#general-changes

**Related known issues with NPM:**

- 5.0.2 Cannot find module '@rollup/rollup-linux-arm64-gnu' on multi platform build [#15167](https://github.com/vitejs/vite/issues/15167)
- Cannot find module @rollup/rollup-win32-x64-msvc on a fresh created js vanilla project [#17284](https://github.com/vitejs/vite/discussions/17284)

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -106,6 +106,9 @@
     "node": ">=22"
   },
   "overrides": {
+    "vite": {
+      "rollup": "npm:@rollup/wasm-node@^4.30.1"
+    },
     "vitest": "^3.0.0-beta.4"
   }
 }
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -96,6 +96,8 @@
     },
     "node_modules/@alloc/quick-lru": {
       "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@alloc/quick-lru/-/quick-lru-5.2.0.tgz",
+      "integrity": "sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==",
       "dev": true,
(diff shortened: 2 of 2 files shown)


frontend/package.json
```diff
@@ -106,6 +106,9 @@
     "node": ">=22"
   },
   "overrides": {
+    "vite": {
+      "rollup": "npm:@rollup/wasm-node"
```
gregory-j-baker, line 110:
```suggestion
      "rollup": "@rollup/wasm-node"
```

**Reason:** the `npm:` prefix isn't needed since we're already using npm.

sebastien-comeau: https://github.com/vitejs/vite/issues/15167#issuecomment-1833274104

gregory-j-baker: Works on my end.
Anyway, shouldn't you also add a version so we have reproducible builds?

sebastien-comeau: version specified now

gregory-j-baker (approved):

PR #89, opened 2025-01-15
fix(frontend): fix broken route path comparison

gregory-j-baker:
`findRouteByPath(..)` would throw when a trailing slash was encountered


sebastien-comeau (approved):
Consider writing unit tests if they don't exists yet.

PR #96, opened 2025-01-16
first name screen

faiza-jahanzeb:
Added new page for "in person case" with input field and button (non functional)
Added navigation to and from "In person case" page in menu.
image


frontend/app/routes/protected/person-case/new-first-name.tsx
gregory-j-baker, line 1: I think you should move this component inside of `first-name.tsx`. In my opinion, there's no benefit to putting this in its own file.

Also, the route handle won't function as expected from inside this file, as they route handles are only considered for routes (not components). Your component works because the route that's using it just happens to load the same i18n namespaces.

Finally, change `<form>` to `<Form>` (imported from react-router). It functions pretty much the same but allows for clientside form posting.

gregory-j-baker (approved):

PR #134, opened 2025-01-27
feat(frontend): Progress Component

Dario-Au:
## Summary

Added Progress Component

## Types of changes

## Checklist


frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,59 @@
...
+  default: 'bg-gray-600',
+  green: 'bg-green-600',
+  red: 'bg-red-500',
+  yellow: 'bg-yellow-400',
+};
+
+const rootBaseClassName = 'relative w-full overflow-hidden rounded-full bg-gray-200';
+const indicatorBaseClassName = 'h-full w-full flex-1 transition-all';
+
+export interface ProgressProps extends React.ComponentPropsWithoutRef<typeof ProgressPrimitive.Root> {
```
gregory-j-baker, line 28: Not sure if `ComponentPropsWithoutRef` is the correct type to use here since React v19. @sebastien-comeau do you know?

frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,59 @@
...
+  value: number;
+}
+
+export function Progress({ className, size = 'base', variant = 'default', value, label, ...props }: ProgressProps) {
+  const { currentLanguage } = useLanguage();
+
+  return (
+    <>
+      {label && currentLanguage && (
+        <p id="progress-label" className="mb-2">{`${label} ${formatPercent(value, currentLanguage)}`}</p>
```
gregory-j-baker, line 42: If more than one progress component is added to a page, the `id` attribute will not be unique.. so you need to allow an ID to be passed in so if we do need multiple components on a page, we can specify their IDs externally.

frontend/app/components/progress.tsx
sebastien-comeau: > React ComponentPropsWithoutRef Explained
> Definition and Usage
> ComponentPropsWithoutRef is a react function component that can type their props while excluding the ref attribute. This is particularly useful to ensure an element does not receive a ref as part of its props.
> 
> When to Use ComponentPropsWithoutRef
> You should consider using ComponentPropsWithoutRef when creating a component that wraps an HTML element but should not expose a ref to its parent components. This helps maintain a clear component definition and prevents misuse of the ref.

https://www.dhiwise.com/post/understanding-the-role-of-componentpropswithoutref-in-react

frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,59 @@
...
+      )}
+      <ProgressPrimitive.Root
+        className={cn(rootBaseClassName, sizes[size], className)}
+        data-testid="progress-root"
```
gregory-j-baker, line 46: The same reasoning applies to `data-testid` as `id` (https://github.com/DTS-STN/future-sir/pull/134/files#r1931001559).

frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,59 @@
...
+        value={value}
+        {...props}
+        aria-labelledby={label && 'progress-label'}
+      >
+        <ProgressPrimitive.Indicator
+          className={cn(indicatorBaseClassName, variants[variant])}
+          style={{ transform: `translateX(-${100 - value}%)` }}
+          data-testid="progress-indicator"
```
gregory-j-baker, line 54: The same reasoning applies to `data-testid` as `id` (https://github.com/DTS-STN/future-sir/pull/134/files#r1931001559).

frontend/app/components/progress.tsx
sebastien-comeau: `data-testid` and `id` here are since `...props` spreading will override them.

frontend/app/utils/string-utils.ts
```diff
@@ -5,3 +5,13 @@ export function randomString(len: number, allowedChars = '0123456789abcdefghijkl
...
   return Array(len).fill(undefined).map(toRandomChar).join('');
 }
+
+/**
+ * @param value - The whole number percentage (i.e. multiplied by 100) to be formatted
+ * @param locale - The Canadian locale to be used for formatting
+ * @returns - The number formatted as a percentage in the givenlocale
+ */
+export function formatPercent(value: number, locale: Language) {
+  if (!/en|fr/i.test(locale)) throw new Error(`locale must be 'en' or 'fr'`);
```
gregory-j-baker, line 15: `locale` will always be `en` or `fr` because of https://github.com/DTS-STN/future-sir/blob/main/frontend/app/%40types/global.d.ts#L19

Also, I think `locale` should be renamed to match `language` to match the type.

frontend/app/components/progress.tsx
gregory-j-baker: But in this case, you might want access to the ref (for example, if you want to focus or scroll to it).

frontend/app/utils/string-utils.ts
sebastien-comeau: "`always`" if you don't pass `any` as `locale` argument or call it in Javascript where Typescript doesn't matter.

frontend/app/components/progress.tsx
gregory-j-baker: Gotcha. I missed that point.

frontend/app/components/progress.tsx
sebastien-comeau: I guess, why not.

gregory-j-baker: So in this case it should just be `ComponentProps`.

frontend/app/utils/string-utils.ts
gregory-j-baker: Sure. If you want to be pedantic. Maybe we also check that `value` isn't a URL or a JSX.Element.

For all intents and purposes, the check can be removed.

gregory-j-baker: Also.. @Dario-Au, can you add the return type to the function? We've been trying to be explicit about that to enhance readability.

frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,68 @@
...
+const indicatorBaseClassName = 'h-full w-full flex-1 transition-all';
+
+export interface ProgressProps extends React.ComponentProps<typeof ProgressPrimitive.Root> {
+  className?: string;
+  id?: string;
+  size?: keyof typeof sizes;
+  variant?: keyof typeof variants;
+  label: string;
+  value: number;
+  testid?: string;
```
gregory-j-baker, line 35: `data-testid` is already a prop of `ProgressPrimitive.Root`, so you don't need to specify it here.

frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export function Progress({
+  className,
+  id = 'progress-label',
+  size = 'base',
+  variant = 'default',
+  value,
+  label,
+  testid = 'progress-indicator',
```
gregory-j-baker, line 45: This can be removed.

frontend/app/components/progress.tsx
```diff
@@ -0,0 +1,68 @@
...
+        className={cn(rootBaseClassName, sizes[size], className)}
+        data-testid="progress-root"
+        value={value}
+        {...props}
+        aria-labelledby={label && id}
+      >
+        <ProgressPrimitive.Indicator
+          className={cn(indicatorBaseClassName, variants[variant])}
+          style={{ transform: `translateX(-${100 - value}%)` }}
+          data-testid={testid}
```
gregory-j-baker, line 63: I don't think this is needed since there is a testid on the parent that can be used to isolate this component during testing.

gregory-j-baker (approved):

PR #135, opened 2025-01-27
(frontend): Add input select and input option component

faiza-jahanzeb:
## Summary

Add input select and input option component

## Types of changes

## Checklist


frontend/app/components/input-option.tsx
gregory-j-baker, line 1: I don't see why we need this component. It does nothing but add a default `data-testid` attribute to an `<option>`.

I feel that this file just clutters the app.

faiza-jahanzeb: added select component that will use this option component

gregory-j-baker: Now that you've added `input-select.tsx`, I think you should move this component into that file. There's no good reason for it to be in its own file, it just creates clutter.

faiza-jahanzeb: this option component will be used in date picker, that's why it's in separate file..

frontend/app/components/input-select.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const inputHelpMessageId = `input-${id}-help`;
+  const inputLabelId = `input-${id}-label`;
+  const inputTestId = `input-${id}-test`;
+  const inputWrapperId = `input-${id}`;
+
+  function getAriaDescribedby() {
+    const ariaDescribedby = [];
+    if (helpMessage) ariaDescribedby.push(inputHelpMessageId);
+    return ariaDescribedby.length > 0 ? ariaDescribedby.join(' ') : undefined;
+  }
```
gregory-j-baker, line 41: This function can be replaced with a ternary in the component.

ex:

``` jsx
<select aria-describedby={helpMessage && inputHelpMessageId} .../>
```

frontend/app/components/input-option.tsx
gregory-j-baker: Thinking more about this, I still don't think this component is needed at all. It does nothing except add a `data-testid` attribute to `<option>`.

I think this component should be removed and anywhere that it is used should be replaced with `<option>`.

There's no need to clutter the application with unnecessary components.

frontend/tests/components/input-select.test.tsx
```diff
@@ -0,0 +1,83 @@
+import { render } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+import { InputSelect } from '~/components/input-select';
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+  }),
+}));
```
gregory-j-baker, line 10: This can be removed because of <https://github.com/DTS-STN/future-sir/blob/main/frontend/tests/setup.ts#L5>.

frontend/app/components/input-option.tsx
sebastien-comeau: Looking at his usage in CDCP and I agree with Greg.

faiza-jahanzeb: updated

frontend/app/components/input-select.tsx
```diff
@@ -0,0 +1,81 @@
...
+        {...restInputProps}
+      >
+        {options.map((optionProps, index) => {
+          const inputOptionId = `${id}-select-option-${index}`;
+          return (
+            <option
+              id={inputOptionId}
+              data-testid={inputOptionId}
+              key={optionProps.key}
+              value={optionProps.value}
```
gregory-j-baker, line 66: These two attributes are not needed since they are included in `{...optionProps}`.

frontend/app/components/input-select.tsx
```diff
@@ -0,0 +1,81 @@
...
+              {...optionProps}
```
gregory-j-baker, line 67: `{...optionProps}` includes `children`, so you either need to spread this into its own variable to remove it from `{...optionProps}`, or remove `{optionProps.children}` below.

frontend/app/components/input-select.tsx
gregory-j-baker: From your other comment:

> key is required for option

Maybe I'm misunderstanding. Since `{...optionProps}` already contains a `key` key/value, then it will be included in the component props automatically when you use `{...optionProps}`, right? So explicitly setting it is redundant.

gregory-j-baker (approved):

PR #149, opened 2025-01-28
(frontend): Add component for date picker field

faiza-jahanzeb:
## Summary

Add component for date picker field

## Types of changes

## Checklist

## Screenshots

image

### DIFF
--- a/frontend/app/components/date-picker-field.tsx
+++ b/frontend/app/components/date-picker-field.tsx
@@ -1,0 +1,425 @@
+import { Fragment } from 'react';
+import type { JSX, ReactNode } from 'react';
+
+import { useTranslation } from 'react-i18next';
+
+import { InputError } from '~/components/input-error';
+import { InputHelp } from '~/components/input-help';
+import { InputLabel } from '~/components/input-label';
+import { InputLegend } from '~/components/input-legend';
+import { useLanguage } from '~/hooks/use-language';
+import { extractDateParts, getLocalizedMonths } from '~/utils/date-utils';
+import { padWithZero } from '~/utils/string-utils';
+import { cn } from '~/utils/tailwind-utils';
+
+type DatePart = 'year' | 'month' | 'day';
+
+type AriaErrorMessage = Readonly<Record<DatePart, string | undefined>>;
+
+/**
+ * Order of date parts based on language
+ */
+const DATE_PART_ORDER = {
+  en: ['month', 'day', 'year'] as const,
+  fr: ['day', 'month', 'year'] as const,
+} as const satisfies Record<Language, DatePart[]>;
+
+const inputStyles = {
+  base: 'rounded-lg border-gray-500 focus:border-blue-500 focus:ring-blue-500 block focus:ring focus:outline-none',
+  disabled: 'disabled:bg-gray-100 disabled:pointer-events-none disabled:cursor-not-allowed disabled:opacity-70',
+  error: 'border-red-500 focus:border-red-500 focus:ring-red-500',
+} as const;
+
+/**
+ * Props for the DatePickerField component
+ */
+export interface DatePickerFieldProps {
+  defaultValue: string;
+  disabled?: boolean;
+  errorMessages?: {
+    all?: string;
+    day?: string;
+    month?: string;
+    year?: string;
+  };
+  helpMessagePrimary?: React.ReactNode;
+  helpMessagePrimaryClassName?: string;
+  helpMessageSecondary?: React.ReactNode;
+  helpMessageSecondaryClassName?: string;
+  id: string;
+  legend: ReactNode;
+  names: {
+    day: string;
+    month: string;
+    year: string;
+  };
+  required?: boolean;
+}
+
+/**
+ * DatePickerField component
+ *
+ * @param props - Props for the DatePickerField component
+ * @returns JSX.Element
+ */
+export const DatePickerField = ({
+  defaultValue,
+  disabled,
+  errorMessages,
+  helpMessagePrimary,
+  helpMessagePrimaryClassName,
+  helpMessageSecondary,
+  helpMessageSecondaryClassName,
+  id,
+  legend,
+  names,
+  required,
+}: DatePickerFieldProps): JSX.Element => {
+  const { t } = useTranslation(['gcweb']);
+  const { currentLanguage = 'en' } = useLanguage(); // english by default
+  const currentDatePartOrder = DATE_PART_ORDER[currentLanguage];
+
+  // Generate unique IDs for accessibility
+  const baseId = `date-picker-${id}`;
+  const ids = {
+    wrapper: baseId,
+    legend: `${baseId}-legend`,
+    error: {
+      all: `${baseId}-error-all`,
+      month: `${baseId}-error-month`,
+      day: `${baseId}-error-day`,
+      year: `${baseId}-error-year`,
+    },
+    help: {
+      primary: `${baseId}-help-primary`,
+      secondary: `${baseId}-help-secondary`,
+    },
+  };
+
+  // Combine IDs for aria-describedby attribute
+  const ariaDescribedBy: string = [
+    ids.legend, //
+    helpMessagePrimary ? ids.help.primary : false,
+    helpMessageSecondary ? ids.help.secondary : false,
+  ]
+    .filter(Boolean)
+    .join(' ');
+
+  // Generate error messages for each date part
+  const ariaErrorMessage: AriaErrorMessage = currentDatePartOrder.reduce((acc, datePart) => {
+    const errors = [
+      errorMessages?.all ? ids.error.all : false, //
+      errorMessages?.[datePart] ? ids.error[datePart] : false,
+    ]
+      .filter(Boolean)
+      .join(' ');
+
+    return { ...acc, [datePart]: errors || undefined };
+  }, {} as AriaErrorMessage);
+
+  // Extract default date parts from the default value
+  const { day = '', month = '', year = '' } = extractDateParts(defaultValue);
+
+  // Define date picker part fields
+  const datePickerPartFields = {
+    year: (
+      <DatePickerYearField
+        id={id}
+        defaultValue={year}
+        name={names.year}
+        label={t('gcweb:date-picker.year.label')}
+        className="w-full sm:w-32"
+        ariaDescribedBy={ariaDescribedBy}
+        ariaErrorMessage={ariaErrorMessage.year}
+        required={required}
+        disabled={disabled}
+      />
+    ),
+    month: (
+      <DatePickerMonthField
+        id={id}
+        defaultValue={month}
+        name={names.month}
+        label={t('gcweb:date-picker.month.label')}
+        placeholder={t('gcweb:date-picker.month.placeholder')}
+        className="w-full sm:w-auto"
+        currentLanguage={currentLanguage}
+        ariaDescribedBy={ariaDescribedBy}
+        ariaErrorMessage={ariaErrorMessage.month}
+        required={required}
+        disabled={disabled}
+      />
+    ),
+    day: (
+      <DatePickerDayField
+        id={id}
+        defaultValue={day}
+        name={names.day}
+        label={t('gcweb:date-picker.day.label')}
+        className="w-full sm:w-20"
+        ariaDescribedBy={ariaDescribedBy}
+        ariaErrorMessage={ariaErrorMessage.day}
+        required={required}
+        disabled={disabled}
+      />
+    ),
+  } as const satisfies Record<DatePart, JSX.Element>;
+
+  return (
+    <div id={ids.wrapper}>
+      <fieldset className="space-y-2">
+        <InputLegend id={ids.legend}>{legend}</InputLegend>
+
+        {/* Error Messages */}
+        {errorMessages && Object.values(errorMessages).filter(Boolean).length > 0 && (
+          <div className="space-y-2">
+            {errorMessages.all && <InputError id={ids.error.all}>{errorMessages.all}</InputError>}
+            {currentDatePartOrder
+              .filter((datePart) => Boolean(errorMessages[datePart]))
+              .map((datePart) => (
+                <p key={datePart}>
+                  <InputError id={ids.error[datePart]}>{errorMessages[datePart]}</InputError>
+                </p>
+              ))}
+          </div>
+        )}
+
+        {/* Help Messages - Primary */}
+        {helpMessagePrimary && (
+          <InputHelp id={ids.help.primary} className={helpMessagePrimaryClassName}>
+            {helpMessagePrimary}
+          </InputHelp>
+        )}
+
+        {/* Date Picker Part Fields */}
+        <div className="flex flex-col space-y-2 sm:flex-row sm:space-y-0 sm:space-x-2">
+          {currentDatePartOrder.map((datePart) => (
+            <Fragment key={datePart}>{datePickerPartFields[datePart]}</Fragment>
+          ))}
+        </div>
+
+        {/* Help Messages - Secondary */}
+        {helpMessageSecondary && (
+          <InputHelp id={ids.help.secondary} className={helpMessageSecondaryClassName}>
+            {helpMessageSecondary}
+          </InputHelp>
+        )}
+      </fieldset>
+    </div>
+  );
+};
+
+/**
+ * Props for the DatePickerMonthField component
+ */
+interface DatePickerMonthFieldProps {
+  ariaDescribedBy: string;
+  ariaErrorMessage?: string;
+  className?: string;
+  currentLanguage?: Language;
+  defaultValue: string;
+  disabled?: boolean;
+  id: string;
+  label: string;
+  name: string;
+  placeholder: string;
+  required?: boolean;
+}
+
+/**
+ * DatePickerMonthField component
+ *
+ * @param props - Props for the DatePickerMonthField component
+ * @returns JSX.Element
+ */
+function DatePickerMonthField({
+  ariaDescribedBy,
+  ariaErrorMessage,
+  className,
+  currentLanguage,
+  defaultValue,
+  disabled,
+  id,
+  label,
+  name,
+  placeholder,
+  required,
+}: DatePickerMonthFieldProps): JSX.Element {
+  const baseId = `date-picker-${id}-month`;
+  const ids = {
+    wrapper: baseId,
+    label: `${baseId}-label`,
+    select: `${baseId}-select`,
+    optionUnselected: `${baseId}-option-unselected`,
+    option: (monthIndex: number) => `${baseId}-option-${monthIndex}`,
+  };
+  const hasErrorMessage = !!ariaErrorMessage;
+  const months = getLocalizedMonths(currentLanguage ?? 'en');
+
+  return (
+    <div id={ids.wrapper} className="space-y-2">
+      <InputLabel id={ids.label} htmlFor={ids.select}>
+        {label}
+      </InputLabel>
+      <select
+        aria-describedby={ariaDescribedBy}
+        aria-errormessage={ariaErrorMessage}
+        aria-invalid={hasErrorMessage}
+        aria-labelledby={ids.label}
+        aria-required={required}
+        className={cn(inputStyles.base, inputStyles.disabled, hasErrorMessage && inputStyles.error, className)}
+        defaultValue={defaultValue}
+        disabled={disabled}
+        id={ids.select}
+        name={name}
+        required={required}
+      >
+        <option id={ids.optionUnselected} value="" disabled hidden>
+          {placeholder}
+        </option>
+        {months.map((month) => {
+          return (
+            <option id={ids.option(month.index)} key={month.index} value={padWithZero(month.index, 2)}>
+              {month.text}
+            </option>
+          );
+        })}
+      </select>
+    </div>
+  );
+}
+
+/**
+ * Props for the DatePickerYearField component
+ */
+interface DatePickerYearFieldProps {
+  ariaDescribedBy: string;
+  ariaErrorMessage?: string;
+  className?: string;
+  defaultValue: string;
+  disabled?: boolean;
+  id: string;
+  label: string;
+  name: string;
+  required?: boolean;
+}
+
+/**
+ * DatePickerYearField component
+ *
+ * @param props - Props for the DatePickerYearField component
+ * @returns JSX.Element
+ */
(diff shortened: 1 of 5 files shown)


frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+  const inputHelpMessagePrimaryId = `${inputWrapperId}-help-primary`;
+  const inputHelpMessageSecondaryId = `${inputWrapperId}-help-secondary`;
+  const inputLegendId = `${inputWrapperId}-legend`;
+
+  const getAriaDescribedBy = useCallback(() => {
+    const ariaDescribedby = [inputLegendId];
+    if (helpMessagePrimary) ariaDescribedby.push(inputHelpMessagePrimaryId);
+    if (helpMessageSecondary) ariaDescribedby.push(inputHelpMessageSecondaryId);
+    return ariaDescribedby.join(' ');
+  }, [helpMessagePrimary, helpMessageSecondary, inputHelpMessagePrimaryId, inputHelpMessageSecondaryId, inputLegendId]);
```
gregory-j-baker, line 76: Is `useCallback()` needed here? I don't quite get the reason for it.

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+
+  const getAriaErrorMessageYear = useCallback(() => {
+    const ariaDescribedby = [];
+    if (errorMessages?.all) ariaDescribedby.push(inputErrorIdAll);
+    if (errorMessages?.year) ariaDescribedby.push(inputErrorIdMonth);
+    return ariaDescribedby.length > 0 ? ariaDescribedby.join(' ') : undefined;
+  }, [errorMessages?.all, errorMessages?.year, inputErrorIdAll, inputErrorIdMonth]);
```
gregory-j-baker, line 83: Is `useCallback()` needed here?

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+        label={t('gcweb:date-picker.year.label')}
+        className="w-full sm:w-32"
+        ariaDescribedBy={getAriaDescribedBy()}
+        ariaErrorMessage={getAriaErrorMessageYear()}
+        required={required}
+        disabled={disabled}
+      />
+    ),
+    [disabled, getAriaDescribedBy, getAriaErrorMessageYear, id, names.year, required, t, value.year],
+  );
```
gregory-j-baker, line 100: Is `useMemo()` needed here?

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+
+  const getAriaErrorMessageMonth = useCallback(() => {
+    const ariaDescribedby = [];
+    if (errorMessages?.all) ariaDescribedby.push(inputErrorIdAll);
+    if (errorMessages?.month) ariaDescribedby.push(inputErrorIdMonth);
+    return ariaDescribedby.length > 0 ? ariaDescribedby.join(' ') : undefined;
+  }, [errorMessages?.all, errorMessages?.month, inputErrorIdAll, inputErrorIdMonth]);
```
gregory-j-baker, line 107: is `useCallback()` needed here?

frontend/app/components/date-picker-field.tsx
gregory-j-baker, line 1: Why all the `useCallback()` and `useMemo()` calls? This feels like a code smell to me.

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+                {datePickerErrorMessages.year && <p>{datePickerErrorMessages.year}</p>}
+              </>
+            )}
+          </div>
+        )}
+        {helpMessagePrimary && (
+          <InputHelp
+            id={inputHelpMessagePrimaryId}
+            className={cn('mb-2', helpMessagePrimaryClassName)}
+            data-testid="date-picker-help-primary"
```
gregory-j-baker, line 220: This will collide with other `data-testid` attributes if multiple date pickers are rendered in the same document.

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+              {datePickerDay}
+              {datePickerYear}
+            </>
+          )}
+        </div>
+        {helpMessageSecondary && (
+          <InputHelp
+            id={inputHelpMessageSecondaryId}
+            className={cn('mt-2', helpMessageSecondaryClassName)}
+            data-testid="date-picker-help-secondary"
```
gregory-j-baker, line 244: This will collide with other `data-testid` attributes if multiple date pickers are rendered in the same document.

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+        aria-invalid={!!ariaErrorMessage}
+        aria-labelledby={inputLabelId}
+        aria-required={required}
+        className={cn(inputBaseClassName, inputDisabledClassName, ariaErrorMessage && inputErrorClassName, className)}
+        data-testid="date-picker-month-select"
+        defaultValue={defaultValue}
+        disabled={disabled}
+        id={selectId}
+        name={name}
+        required={required}
```
gregory-j-baker, line 304: This could be simplified by moving some of these attributes into a `{...props}` spread.

frontend/app/components/date-picker-field.tsx
```diff
@@ -0,0 +1,428 @@
...
+  const { currentLanguage } = useLanguage();
+  const months = useMonths(currentLanguage ?? 'en');
+
+  const selectId = `date-picker-${id}-month`;
+  const wrapperId = `date-picker-${id}-month-wrapper`;
+  const inputLabelId = `date-picker-${id}-month-label`;
+  const inputOptionUnselectedId = `date-picker-${id}-month-option-unselected`;
+
+  return (
+    <div id={wrapperId} data-testid="date-picker-month">
```
gregory-j-baker, line 288: `data-testid` will collide when multiple date pickers are rendered in a document

frontend/app/utils/string-utils.ts
```diff
@@ -14,3 +14,16 @@ export function randomString(len: number, allowedChars = '0123456789abcdefghijkl
...
+ * If the value is not a number or its length is greater than or equal to maxLength, it returns the value as a string.
+ * @param value - The number to pad.
+ * @param maxLength - The maximum length of the resulting string.
+ * @returns The padded string or the value as a string if it's not a number or its length is already greater than or equal to maxLength.
+ */
+export const padWithZero = (value: number, maxLength: number) => {
+  if (Number.isNaN(value)) return value.toString();
+  if (value.toString().length >= maxLength) return value.toString();
+  return value.toString().padStart(maxLength, '0');
+};
```
gregory-j-baker, line 29:
```suggestion
export function padWithZero(value: number, maxLength: number): string {
  if (Number.isNaN(value)) return value.toString();
  if (value.toString().length >= maxLength) return value.toString();
  return value.toString().padStart(maxLength, '0');
}
```

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/**
+ * Custom hook to retrieve an array of months based on the provided locale and format.
+ * @param locale - The locale to use for formatting the months.
+ * @param format - The format for displaying the months.
+ * @returns An array containing objects with month index and formatted month text.
+ */
+export function useMonths(locale: string, format: 'numeric' | '2-digit' | 'long' | 'short' | 'narrow' | undefined = 'long') {
+  return useMemo(() => {
```
gregory-j-baker, line 49: I don't think this should `useMemo()`. If the calling component needs to memoize it, then it should do so there, not in this function.

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  return date.getFullYear() === year && date.getMonth() === month && date.getDate() === day;
+}
+
+/**
+ * Custom hook to retrieve an array of months based on the provided locale and format.
+ * @param locale - The locale to use for formatting the months.
+ * @param format - The format for displaying the months.
+ * @returns An array containing objects with month index and formatted month text.
+ */
+export function useMonths(locale: string, format: 'numeric' | '2-digit' | 'long' | 'short' | 'narrow' | undefined = 'long') {
```
gregory-j-baker, line 48:
```suggestion
export function useMonths(locale: string, format: 'numeric' | '2-digit' | 'long' | 'short' | 'narrow' | undefined = 'long'): { index: number; text: string; }[] {
```

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    return [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12].map((monthIndex) => {
+      const mm = monthIndex < 10 ? `0${monthIndex}` : monthIndex;
+      const date = new Date(`2017-${mm}-01T00:00:00Z`);
+      return {
+        index: monthIndex,
+        text: formatter.format(date),
+      };
+    });
+  }, [format, locale]);
+}
```
gregory-j-baker, line 60:
```suggestion
export function useMonths(locale: string, format: "numeric" | "2-digit" | "long" | "short" | "narrow" | undefined = "long"): { index: number; text: string }[] {
  const formatter = new Intl.DateTimeFormat(locale, { month: format, timeZone: "UTC" });

  return Array.from({ length: 12 }, (_, i) => ({
    index: i + 1, // month index (1-based)
    text: formatter.format(new Date(Date.UTC(2017, i, 1))), // formatted month name
  }));
}
```

I think that works the same, and is more concise. I can't say 100% since there's no test for this function.

sebastien-comeau: The custom hook should be moved into `/app/hooks/use-months.ts`

frontend/app/utils/date-utils.ts
sebastien-comeau: I've done it in CDCP for perfromance. No need to call that factory logic orver and over. We should create a map or months/locale in `useMemo()` hook file instead to have 1 singleton per locale for optimization.

frontend/app/components/date-picker-field.tsx
sebastien-comeau: It's not a bad thing. It depends how the test is written.

frontend/app/utils/date-utils.ts
gregory-j-baker: In my opinion it's premature optimization. This isn't computationally intensive. And even if it was, the memoization should be done outside of the function.

frontend/app/components/date-picker-field.tsx
sebastien-comeau: It's be done that way to store the components and display them in a specific order based on current `language`.

```ts
{currentLanguage === 'fr' ? (
  <>
    {datePickerDay}
    {datePickerMonth}
    {datePickerYear}
  </>
) : (
  <>
    {datePickerMonth}
    {datePickerDay}
    {datePickerYear}
  </>
)}
```

frontend/app/components/date-picker-field.tsx
gregory-j-baker: It's not *necessarily* a bad thing, but if the idea behind `data-testid` is to make it easier to isolate components during testing, then adding a unique ID to these elements only supports that goal.

frontend/app/utils/date-utils.ts
sebastien-comeau: You look like a premature optimization. 🏃

frontend/app/utils/date-utils.ts
gregory-j-baker: Why *is* this a hook? It doesn't need to be.

frontend/app/components/date-picker-field.tsx
sebastien-comeau: I helped me in my unit tests to target those element for assertion. Component unit tests in FSIR is different since you opted to use snapshots. 🤷

frontend/app/components/date-picker-field.tsx
gregory-j-baker: This can't be done without complex memoization?

frontend/app/components/date-picker-field.tsx
gregory-j-baker: Would it not help more to make the test id `date-picker-month-${id}`?

---
Also.. now all I can think about is picking dates. https://www.wikihow.com/Harvest-Dates

frontend/app/utils/date-utils.ts
sebastien-comeau: AI suggestion base on your suggestion and it was my train of thoughts when I created it.

<hr />

Yes, the code provided can indeed be used as a React hook, though it is more like a utility function structured as a hook since it doesn't handle state, side effects, or lifecycle events. Here's some feedback and possible improvements:

**Why It's Suitable as a Hook**

- It uses parameters (locale and format) that allow flexibility.
- Returns reusable, memoized data (formatted month names).

**Suggested Improvement**

**_Memoization_**

Without useMemo, the hook recalculates every time the parent component renders. You can optimize it as follows:

```ts
import { useMemo } from "react";

export function useMonths(
  locale: string,
  format: "numeric" | "2-digit" | "long" | "short" | "narrow" | undefined = "long"
): { index: number; text: string }[] {
  return useMemo(() => {
    const formatter = new Intl.DateTimeFormat(locale, { month: format, timeZone: "UTC" });
    return Array.from({ length: 12 }, (_, i) => ({
      index: i + 1,
      text: formatter.format(new Date(Date.UTC(2017, i, 1))),
    }));
  }, [locale, format]);
}
```

frontend/app/components/date-picker-field.tsx
sebastien-comeau: I don't know, it can be explored it in another PR.

frontend/app/components/date-picker-field.tsx
sebastien-comeau: If using `ByTestId` with a `Regex` expression then it could work.

ken-blanchard: Adding a data-testid attribute to a reusable component isn't inherently bad, even if multiple instances of that component are used on the same page. The key is to ensure that your selectors are scoped appropriately to avoid conflicts if you need to target a specific component.

For example, if you have multiple DatePicker components, you can narrow down the scope of your selector by targeting a specific parent element or using more specific attributes. This way, you can still use data-testid for testing purposes without running into issues with multiple instances.

Adding a unique ID to the data-testid attribute, can help differentiate between multiple instances of the same component. However, it requires knowing the specific ID for targeting, which can add also add complexity to your tests.

gregory-j-baker: Testing a page with multiple date pickers is naturally more complex. I'm not convinced that making the `data-testid` values unique actually adds *even more complexity* on top of that.

Sure.. it's possible to use regex or finely tuned selectors to target specific elements, but using a unique `data-testid` is a simpler and arguably more reliable approach.

Help me understand why you are so against this approach? I only see benefit with no downsides.. what am I missing?

frontend/app/components/date-picker-field.tsx
gregory-j-baker: Why add more complex code now, only to (potentially) remove it later. Just throw the component on a page somewhere without the memoization and see if it works..

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,55 @@
+import { useMemo } from 'react';
+
+import { padWithZero } from './string-utils';
+
+/**
+ * Parses a date string in the format "YYYY-MM-DD" and returns an object with the parsed components.
+ * @param date The date string to parse.
+ * @returns An object containing the parsed components (year, month, day). Returns an empty object if parsing fails or if the date does not exist.
+ */
+
```
gregory-j-baker, line 10: **Potential nitpick:** this blank line should be removed so the jsdoc is "attached" to the function.

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,55 @@
+import { useMemo } from 'react';
+
+import { padWithZero } from './string-utils';
```
gregory-j-baker, line 3: To be consistent with the rest of the application, this should be an aliased (`~/`) import.

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  return {
+    year: padWithZero(year, 4),
+    month: padWithZero(month, 2),
+    day: padWithZero(day, 2),
+  };
+}
+
+function dateExists(year: number, month: number, day: number): boolean {
```
gregory-j-baker, line 37: Should this be exported so it can be called externally?

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,55 @@
+import { useMemo } from 'react';
```
gregory-j-baker, line 1: This should be removed.

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { padWithZero } from './string-utils';
+
+/**
+ * Parses a date string in the format "YYYY-MM-DD" and returns an object with the parsed components.
+ * @param date The date string to parse.
+ * @returns An object containing the parsed components (year, month, day). Returns an empty object if parsing fails or if the date does not exist.
+ */
+
+export function extractDateParts(date: string) {
```
gregory-j-baker, line 11:
```suggestion
export function extractDateParts(date: string): { year?: string; month?: string; day?: string } {
```

frontend/app/utils/date-utils.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!dateExists(year, month - 1, day)) {
+    return {};
+  }
+
+  return {
+    year: padWithZero(year, 4),
+    month: padWithZero(month, 2),
+    day: padWithZero(day, 2),
+  };
```
gregory-j-baker, line 34: I think this function can be tidied up a bit:

``` typescript
const [yearStr, monthStr, dayStr] = date.split('-');

const year = Number(yearStr);
const month = Number(monthStr);
const day = Number(dayStr);

if (isNaN(year) || isNaN(month) || isNaN(day)) return {};
if (!dateExists(year, month - 1, day)) return {};

return {
  year: padWithZero(year, 4),
  month: padWithZero(month, 2),
  day: padWithZero(day, 2),
};
```

frontend/app/components/date-picker-field.tsx
sebastien-comeau: Ok, since `id` is not auto-generated then it make sense.

frontend/app/utils/date-utils.ts
sebastien-comeau: LLM doesn't want to hurt your feelings. Remember that if you're using the return value in `useEffect` or other hooks that need dependencies, it might be a problem since the array will be re-generated will have a different reference. 🤷

gregory-j-baker: Even if that is true, that is of no concern to *this function*.

frontend/app/components/date-picker-field.tsx
faiza-jahanzeb: updated the data test id every where starting with the pattern {`date-picker-${id}`}

frontend/app/utils/date-utils.ts
sebastien-comeau: This function is a custom hook, and hooks can only be called directly within functional components or other hooks. Since it doesn't return a primitive value, returning a memoized result is essential. Based on our discussion about dependency management in hooks like `useEffect`, `useMemo`, and `useCallback`, I believe retaining `useMonths` offers a cleaner and more maintainable approach by encapsulating memoization and dependency handling. This avoids repetitive calls to `getMonths` across the codebase.

```ts
// ~/utilities/date-utils.ts
export function getMonths(locale, format = 'long') {
  const formatter = new Intl.DateTimeFormat(`${locale}-CA`, { month: format, timeZone: "UTC" });

  return Array.from({ length: 12 }, (_, i) => {
    const monthIndex = i + 1;
    const mm = monthIndex.toString().padStart(2, "0");
    const date = new Date(`2017-${mm}-01T00:00:00Z`);
    return {
      index: monthIndex,
      text: formatter.format(date),
    };
  });
}

// ~/hooks/use-months.ts
import { useMemo } from "react";

import { getMonths } form '~/utilities/date-utils'

export function useMonths(locale, format = 'long') {
  return useMemo(() => getMonths(locale, format), [locale, format]);
}
```

faiza-jahanzeb: @sebastien-comeau 
Looking at the date-picker-field, the datePickerMonth is called inside useMemo. If we add the currentLanguage there to get the locale(for getting months), we don't need to add useMemo in use-months, and the date-utils will remain as it is now..

in date-picker-field.tsx

  const datePickerMonth = useMemo(
    () => (
      <DatePickerMonth
        id={id}
        defaultValue={value.month ?? ''}
        name={names.month}
        label={t('gcweb:date-picker.month.label')}
        placeholder={t('gcweb:date-picker.month.placeholder')}
        className="w-full sm:w-auto"
        currentLanguage={currentLanguage}
        ariaDescribedBy={getAriaDescribedBy()}
        ariaErrorMessage={getAriaErrorMessageMonth()}
        required={required}
        disabled={disabled}
      />
    ),
    [currentLanguage, disabled, getAriaDescribedBy, getAriaErrorMessageMonth, id, names.month, required, t, value.month],
  );

if you don't agree with this change then I can revert it back and add the useMemo in useMonth and move it to hooks folder

frontend/app/routes/protected/person-case/first-name.tsx
```diff
@@ -84,6 +85,16 @@ export default function FirstName({ loaderData, actionData, params }: Route.Comp
...
+        <DatePickerField
+          defaultValue={'2025-01-01'}
+          id={'someId'}
+          legend={undefined}
+          names={{
+            day: 'day',
+            month: 'month',
+            year: 'year',
+          }}
+        />
```
faiza-jahanzeb, line 97: I'll revert this change

frontend/app/utils/date-utils.ts
sebastien-comeau: You can remove `useMemo`.

frontend/app/components/date-picker-field.tsx
faiza-jahanzeb: updated

sebastien-comeau (approved):

PR #159, opened 2025-01-30
test(frontend): add tests for date picker field

faiza-jahanzeb:
## Summary

Add tests for date picker field

## Types of changes

## Checklist


frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+        legend="legend date picker field test"
+        names={{
+          day: 'day test',
+          month: 'month test',
+          year: 'year test',
+        }}
+      />,
+    );
+    expect(container.innerHTML).toMatchSnapshot('expected html');
+  });
```
gregory-j-baker, line 29: While this works, I think a better approach would be this:

``` tsx
  it('should render date picker field component', () => {
    const RoutesStub = createRoutesStub([
      {
        path: '/',
        Component: () => (
          <DatePickerField
            defaultValue={''}
            id="id"
            legend="legend date picker field test"
            names={{ day: 'day test', month: 'month test', year: 'year test' }}
          />
        ),
      },
    ]);

    const { container } = render(<RoutesStub />);

    expect(container.innerHTML).toMatchSnapshot('expected html');
  });
```

The reason is that using a `<RoutesStub>` tests your component within the context of React Router, which (in my opinion) provides a little more assurances that it works as expected.

It could be argued that wrapping the component in a `<RoutesStub>` is not a *true* unit test (which is a valid point), but I still think we should.

The same comment applies to all of the tests in this file.

sebastien-comeau: The component itself has nothing to do with React Router. I don't think we should add an extra layer of testing over it.

gregory-j-baker: I uses RR hooks, so it does interface with RR.. hence my suggestion.

sebastien-comeau: Fair, I never realised it.

faiza-jahanzeb: updated my pr

faiza-jahanzeb: updated

frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+import { DatePickerField } from '~/components/date-picker-field';
+
+vi.mock(import('react-router'), async (importOriginal) => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    useLocation: vi.fn(),
+  };
+});
```
gregory-j-baker, line 14: This shouldn't be required when using `createRoutesStub()`.

frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+describe('DatePickerField', () => {
+  vi.mocked(useLocation, { partial: true }).mockReturnValue({ pathname: '/en/some-path' });
```
gregory-j-baker, line 17: This shouldn't be required. If you need `useLocation()` to return a path with a language, just set it in the `<RoutesStub>`

gregory-j-baker: I just ran your tests without the `vi.mock()` and `vi.mocked()` and they pass.

frontend/tests/components/__snapshots__/date-picker-field.test.tsx.snap
```diff
@@ -0,0 +1,11 @@
...
+
+exports[`DatePickerField > should render date picker field component with help message primary and help message secondary > expected html 1`] = `"<h2>Unexpected Application Error!</h2><h3 style="font-style: italic;">404 Not Found</h3><p>💿 Hey developer 👋</p><p>You can provide a way better UX than this when your app throws errors by providing your own <code style="padding: 2px 4px; background-color: rgba(200, 200, 200, 0.5);">ErrorBoundary</code> or <code style="padding: 2px 4px; background-color: rgba(200, 200, 200, 0.5);">errorElement</code> prop on your route.</p>"`;
+
+exports[`DatePickerField > should render date picker field component with required > expected html 1`] = `"<h2>Unexpected Application Error!</h2><h3 style="font-style: italic;">404 Not Found</h3><p>💿 Hey developer 👋</p><p>You can provide a way better UX than this when your app throws errors by providing your own <code style="padding: 2px 4px; background-color: rgba(200, 200, 200, 0.5);">ErrorBoundary</code> or <code style="padding: 2px 4px; background-color: rgba(200, 200, 200, 0.5);">errorElement</code> prop on your route.</p>"`;
```
gregory-j-baker, line 11: These are error boundaries. This is not what the component should render.

frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const { container } = render(<RoutesStub />);
+
+    expect(container.innerHTML).toMatchSnapshot('expected html');
+  });
+
+  it('should render date picker field component with required', () => {
+    const RoutesStub = createRoutesStub([
+      {
+        path: '/fr/some-path',
```
gregory-j-baker, line 32: Change this to `path: '/'` since you are not specifying an initial entry when rendering the `<RouteStub>` component below. This will fix your error boundary issue in the snapshot files.

frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const { container } = render(<RoutesStub />);
+
+    expect(container.innerHTML).toMatchSnapshot('expected html');
+  });
+
+  it('should render date picker field component with help message primary and help message secondary', () => {
+    const RoutesStub = createRoutesStub([
+      {
+        path: '/en/some-path',
```
gregory-j-baker, line 57: Change this to `path: '/'` since you are not specifying an initial entry when rendering the `<RouteStub>` component below. This will fix your error boundary issue in the snapshot files.

frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const { container } = render(<RoutesStub />);
+
+    expect(container.innerHTML).toMatchSnapshot('expected html');
+  });
+
+  it('should render date picker field component with default value', () => {
+    const RoutesStub = createRoutesStub([
+      {
+        path: '/en/some-path',
```
gregory-j-baker, line 83: Change this to `path: '/'` since you are not specifying an initial entry when rendering the `<RouteStub>` component below. This will fix your error boundary issue in the snapshot files.

frontend/tests/components/date-picker-field.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const { container } = render(<RoutesStub />);
+
+    expect(container.innerHTML).toMatchSnapshot('expected html');
+  });
+
+  it('should render date picker field component with error message', () => {
+    const RoutesStub = createRoutesStub([
+      {
+        path: '/en/some-path',
```
gregory-j-baker, line 107: Change this to `path: '/'` since you are not specifying an initial entry when rendering the `<RouteStub>` component below. This will fix your error boundary issue in the snapshot files.

sebastien-comeau (approved):

PR #163, opened 2025-01-30
refactor(frontend): Landing Page Refactor

Dario-Au:
## Summary

Work Item: [14941](https://dev.azure.com/ESDCCM/FSIR/_workitems/edit/14941)
Updated landing page to match the mockup

## Types of changes

## Checklist

<details>

  ``` shell
  npm run lint:check
  npm run format:check
  ```
</details>

<details>

  ``` shell
  npm run test
  npm run test:e2e
  ```
</details>

## Additional Notes

- Moved the landing page's ButtonLinks to its own component and added tests for it

## Screenshots

image


frontend/app/components/navigation-button.tsx
sebastien-comeau, line 1: Create a `card` component similar to GC Design System [card](https://design-system.alpha.canada.ca/en/components/card/code/#get-your-code). Don't another existing component like `ButtonLink`, because you have to override almost all of the classes. The only existing you would have to use is `AppLink` for the anchor link.

frontend/app/components/button-link.tsx
```diff
@@ -32,7 +32,7 @@ type ButtonLinkStyleProps = {
   variant?: keyof typeof variants;
 };
 
-type ButtonLinkProps = ComponentProps<typeof AppLink> &
+export type ButtonLinkProps = ComponentProps<typeof AppLink> &
```
gregory-j-baker, line 35: Nitpick: this isn't *bad*, but you could also just do `ComponentProps<ButtonLink>` wherever you need to use this (like we do elsewhere).. just sayin.

Will resolve this comment because it truly is a nitpick.

frontend/app/components/navigation-button.tsx
sebastien-comeau: Using `ButtonLink` comes with default classes that has nothing to do with a `card` link.

gregory-j-baker:
Feel free to remove the `i18n resources loaded from` text from this page. It was just placed there for demo purposes.

frontend/app/components/navigation-button.tsx
```diff
@@ -0,0 +1,48 @@
+import type { ComponentProps } from 'react';
+
+import type { FlipProp, IconProp } from '@fortawesome/fontawesome-svg-core';
+import { faChevronRight } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import { AppLink } from './links';
```
gregory-j-baker, line 7: This should be an aliased import (`~/`). I'm really considering writing an eslint plugin that checks for this.

gregory-j-baker: And I acknowledge that vscode is pretty terrible for this.. I often have to force push my PRs with fixes to the imports.

sebastien-comeau:
`NavigationButton` might be a bit misleading since it's a card acting as a link. I'd suggest renaming it to `CardLink` for clarity. Additionally, the current implementation isn't responsive due to its fixed size.  

Here are some improvement ideas:  

- Add a `tag` (string) property for an optional tag display at the top left.  
- Include `image` and `imageAlt` properties for an optional image display.  
- Add `titleAs` with a union of element types (`'p'`, headings like `'h1'`, etc.), defaulting to `'h3'`, giving consumers flexibility to set appropriate semantic tags.  
- Use the `children` prop for custom card content under the title, allowing richer content than just a string.  

I've created a proposal you can check out here: [CardLink Proposal](https://github.com/DTS-STN/future-sir/tree/seb/card-link)](https://github.com/DTS-STN/future-sir/tree/seb/card-link).

**Screenshots with all props**

image

**Screenshots without image**

image

**Screenshots without tag**

image

**Screenshots without icon**

image

**Screenshots without disabled="true"**

image

gregory-j-baker:
> I'd suggest renaming it to `CardLink` for clarity.

Why not just name it `<Card>` and allow it optional props to become a link.

`<Card>{* content *}</Card>` ← Not a link
`<Card file="routes/index.tsx">{* content *}</Card>` ← A link!
`<Card to="https://google.ca/">{* content *}</Card>` ← Also a link!

sebastien-comeau:
> > I'd suggest renaming it to `CardLink` for clarity.
> 
> Why not just name it `<Card>` and allow it optional props to become a link.
> 
> `<Card>{* content *}</Card>` ← Not a link `<Card file="routes/index.tsx">{* content *}</Card>` ← A link! `<Card to="https://google.ca/">{* content *}</Card>` ← Also a link!

It could be.

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import { AppLink } from '~/components/links';
+import type { I18nRouteFile } from '~/i18n-routes';
+import { cn } from '~/utils/tailwind-utils';
+
+type OptionalAppLink = Omit<ComponentProps<typeof AppLink>, 'to' | 'lang'> & {
+  to?: To;
+  file?: I18nRouteFile;
+};
```
gregory-j-baker, line 16:
```suggestion
type OptionalAppLink = ComponentProps<typeof AppLink>;
```

**Reason:** `<AppLink>` already contains a `to`, and `file` prop. Omitting `to` and then adding it back via a type union is redundant. Also, as far as I can tell, there's no good reason to omit `lang`.

gregory-j-baker: I see the issue with my suggestion, as typescript gets angry because you can't spread `{...props}` into `<AppLink>`.

Gimme a bit to try something.

gregory-j-baker: Sorry.. what you did seems right. My bad. I'll resolve this comment, but I do have to make another one. on this line.

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+import type { FlipProp, IconProp } from '@fortawesome/fontawesome-svg-core';
+import { faChevronRight } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+import { AppLink } from '~/components/links';
+import type { I18nRouteFile } from '~/i18n-routes';
+import { cn } from '~/utils/tailwind-utils';
+
+type OptionalAppLink = Omit<ComponentProps<typeof AppLink>, 'to' | 'lang'> & {
```
gregory-j-baker, line 13: I don't think you have to omit `lang` here.

gregory-j-baker: Also nitpick: the convention for naming prop types is to add a `Props` suffix (ex: `CardBaseProps`)

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,161 @@
...
+function CardBase(props: OptionalAppLink) {
+  const { to, file, disabled } = props;
+  const baseClassName = `select-none rounded-sm ${disabled ? 'bg-white' : 'bg-blue-900'}`;
+  if (to || file) {
+    return (
+      <div className={baseClassName}>
+        <AppLink {...(props as ComponentProps<typeof AppLink>)} />
+      </div>
+    );
+  } else {
```
gregory-j-baker, line 70: Nitpick, the `else {}` is unnecessary here.

``` typescript
const baseClassName = `select-none rounded-sm ${disabled ? 'bg-white' : 'bg-blue-900'}`;

if (to || file) {
  return // component
}

// rest of code
```

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,161 @@
...
+    return (
+      <div className={baseClassName}>
+        <a
+          role="button"
+          aria-disabled={disabled}
+          tabIndex={0}
+          onClick={(e: MouseEvent<HTMLAnchorElement>) => {
```
gregory-j-baker, line 77: Nitpick: the type can be inferred here since onclick functions are always of type `MouseEvent<HTMLAnchorElement, MouseEvent>`

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,161 @@
...
+            e.preventDefault();
+            props.onClick?.(e);
+          }}
+          onKeyDown={(e: KeyboardEvent<HTMLAnchorElement>) => {
```
gregory-j-baker, line 81: Nitpick: the type can be inferred here because onkeydown functions are always of type `KeyboardEvent<HTMLAnchorElement>`

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,161 @@
...
+                aria-hidden
+                icon={faChevronRight}
+                className="text-lg font-bold text-slate-700 underline group-hover:text-blue-700 group-focus:text-blue-700"
+              />
+            </div>
+            {/* Content - Optional */}
+            {children && <div>{children}</div>}
+          </div>
+        </div>
+      </div>
```
gregory-j-baker, line 158: This might be a spicy take, but...

I feel like this component could be made a little more readable by extracting out the various subcomponents (in the same file...not new files).

something like this:

``` tsx
<>
  <Image />
  <Tag />
  <Icon />
  <Title />
  <Content />
</>
```

It allows you to see the overall structure of the component without requiring a bunch of comments (ie: it's *self documenting*).

I see a lot of components in applications that are quite difficult to read because they are basically:

``` tsx
<>
  {/* 500 lines of tsx, a lot of which is conditionally rendered */}
</>
```

Breaking these down into easier to chew bites can go a long way towards increasing readability.

gregory-j-baker: In this case, if you disagree.. I am okay with that. This component isn't *huge*, although I would say it's more difficult to read than it needs to be.

Feel free to mark this as resolved if you don't want to create the additional  local (in-file) components.

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,182 @@
...
+    </>
+  );
+}
+
+type CardTitleProps = { titleAs: CardProps['titleAs']; title: CardProps['title'] };
+
+function CardTitle({ titleAs: TitleComponent = 'h3', title }: CardTitleProps) {
+  return (
+    <div className="flex items-center gap-2">
+      <TitleComponent className="text-lg font-bold text-slate-700 underline group-hover:text-blue-700 group-focus:text-blue-700">
```
sebastien-comeau, line 166: The following style only make "sense" for a `CardLink`.

```
text-slate-700 underline group-hover:text-blue-700 group-focus:text-blue-700
```

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,182 @@
...
+        {title}
+      </TitleComponent>
+      <FontAwesomeIcon
```
sebastien-comeau, line 169: Chevron icon only make "sense" for a `CardLink`.

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,182 @@
...
+        aria-disabled={disabled}
+        tabIndex={0}
+        onClick={(e) => {
+          e.preventDefault();
+          props.onClick?.(e);
+        }}
+        onKeyDown={(e) => {
+          e.preventDefault();
+          props.onKeyDown?.(e);
+        }}
```
sebastien-comeau, line 114: I don't think you have to define `onClick` and `onKeyDown` since you are spreading those `props`.

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,182 @@
...
+      </div>
+    </CardBase>
+  );
+}
+
+/**
+ * Base component for the Card component
+ * Returns an AppLink when the 'To' or 'File' prop is provided.
+ * Otherwise, returns an Anchor tag that functions as a button.
+ */
```
sebastien-comeau, line 90: I don't think this is right. It's either a `CardLink` when the 'To' or 'File' prop is provided or a plain `Card` without action. The `Card` consumer should be able to pass button or anything he want inside the content (`children`). I think we went too far with this multi-purpose `Card` and should have created `CardLink` for this purpose and create another for the other purspose. @gregory-j-baker that component become way too complicated.

Dario-Au: :moyai:

gregory-j-baker: Wrapping a card in an `<AppLink>` doesn't have to be complicated. We don't need yet another component for this.

The complications arise when trying to turn *everything* into attributes.. icon, image, title, alt, etc... these things should be children of the `<Card>` component that the calling component defines.

A card should display a header, some content, and a footer, and these should be separate *subcomponents* that are defined by the calling component. If a `to` or `file` attribute is passed, it gets wrapped in an `<AppLink>`.

``` tsx
function Card() {
  return isLink
    ? <AppLink {...linkprops}><BaseCard {...cardprops} /></AppLink>
    : <BaseCard {...cardprops} />
}

function BaseCard() {
  // the rest of the f'n owl
}
```

frontend/app/components/card.tsx
gregory-j-baker: I was thinking a bit more about this and maybe for the sake of just getting it done, another separate component would be fine.

frontend/app/components/card.tsx
gregory-j-baker: I was thinking a bit more about this and maybe for the sake of just getting it done, another separate component would be fine.

sebastien-comeau:
@gregory-j-baker @Dario-Au I've refactored `Card` components using [@radix-ui/react-slot](https://www.radix-ui.com/primitives/docs/utilities/slot).

`Card` and `CardTitle` components have `asChild` property. It gives the ability to merges their props onto its immediate child.

Usage example with `AppLink`.

```tsx
<Card asChild>
  <AppLink file="routes/public/index.tsx">
    <CardHeader>
      <CardTitle>Card Title</CardTitle>
      <CardDescription>Test Card</CardDescription>
    </CardHeader>
  </AppLink>
</Card>
```

frontend/app/components/card.tsx
```diff
@@ -0,0 +1,133 @@
...
+export function CardHeader({ className, ...props }: HTMLProps<HTMLDivElement>) {
+  return <div className={cn('flex flex-col space-y-1.5 p-6', className)} {...props} />;
+}
+
+type CardTitleProps = HTMLProps<HTMLHeadingElement> & {
+  /**
+   * Use the asChild prop to change the default rendered element for the one passed as a child, merging their props and behavior.
+   * @default 'h3'
+   */
+  asChild?: boolean;
```
gregory-j-baker, line 46: Does the `@default` make sense here when the property is `boolean`?

gregory-j-baker (approved):

gregory-j-baker:
I think this new implementation is better.

gregory-j-baker (approved):

PR #170, opened 2025-01-31
fix(frontend): update page title component

faiza-jahanzeb:
## Summary

Update page title component to display partial underline using css

## Types of changes

## Checklist

## Screenshots

image

image


frontend/app/components/page-title.tsx
```diff
@@ -2,17 +2,31 @@ import type { ComponentProps } from 'react';
...
+    <>
+      {subTitle && <h2 className="mt-10">{subTitle}</h2>}
+
+      <h1
+        id="wb-cont"
+        tabIndex={-1}
+        className={cn(
+          partialUnderlinedClassName,
+          'font-lato mb-8 text-2xl font-bold text-slate-700 focus-visible:ring-3',
+          !subTitle && 'mt-10',
```
gregory-j-baker, line 23: Instead of using margins for this, would it make sense to wrap the page title in a `<div>` with classname `space-y-10` (or whatever).

That way, the spacing would be applied automatically when there is a subtitle, but not applied when the subtitle is omitted.

gregory-j-baker: Also, I'm wondering if the subtitle should be a `<span>` inside of the `<h1>`?

``` tsx
<h1>
  <span className="...">Optional subheading</span>
  Actual page title
</h1>
```

gregory-j-baker: I took a screenshot of CDCP's pagetitle and provided that and the the HTML that was used to render it to ChatGPT, and this is what it suggested:

image

Obviously, take it with a grain of salt because it's Chat GPT.

sebastien-comeau: Here's my suggestion playing with it locally and see how GC Design System display their red underline. They're using `:after` pseudo-class and it make sense.

```ts
export function PageTitle({ children, className, subTitle, ...props }: PageTitleProps) {
  return (
    <div className="mt-10 mb-8">
      {subTitle && <h2>{subTitle}</h2>}
      <h1
        id="wb-cont"
        tabIndex={-1}
        className={cn(
          'font-lato text-3xl font-bold text-slate-700 focus-visible:ring-3',
          'after:mt-2 after:block after:h-1.5 after:w-18 after:bg-red-700',

          className,
        )}
        {...props}
      >
        {children}
      </h1>
    </div>
  );
}
```

sebastien-comeau: image

faiza-jahanzeb: This is the name of the flow on each page, I thought that too that it's a span (part of h1), but @sebastien-comeau  said it should be <h2>. It's done that way in CDCP project, and I think it's something to do with accessibility audit.

gregory-j-baker: I don't think moving the subheading inside of the `<h1>` is an accessibility issue, but for now I guess leave it as CDCP has it.

gregory-j-baker: Re: partial underline.. do we have to do this? Why not just do what CDCP did and have a full-width underline?

image

gregory-j-baker: If we do go with partial underline, I like @sebastien-comeau's a little more (thicker line with just a little more y-spacing.

sebastien-comeau: Follow the design for this one. I don't think it's bad and it matches GC Design System.

gregory-j-baker: @sebastien-comeau I'll let you resolve the comment since you suggested the updated styles

> Here's my suggestion playing with it locally...

gregory-j-baker (dismissed):

sebastien-comeau (approved):

gregory-j-baker (approved):

PR #189, opened 2025-02-07
fix(frontend): improve xstate navigation

gregory-j-baker:
[AB#15081](https://dev.azure.com/ESDCCM/a4b4f7af-bbd7-4f0d-83ab-a123d0b70fda/_workitems/edit/15081) -- Explore XState for managing wizard screen flow

## Summary

Refactoring to how the xstate stuff works.

- Allowing for target-state loading when hitting a route. This prevents the state machine from desynchronizing with the application URL. When you hit `/my-route` then the correct state for `my-route` is loaded.
- Moving some repeated code into helper functions.
- Strict-typing the state names to ensure some of the helper functions are called correctly.

## Types of changes

- [x] 🐛 **bugfix** -- non-breaking change that fixes an issue
- [x] 💥 **breaking change** -- change that causes existing functionality to no longer work as expected
- [x] 📚 **documentation** -- changes to documentation only

## Checklist

- [x] added adequate logging for new or updated functionality
- [x] documentation has been updated to reflect the changes
- [x] linked this PR to a related issue

### DIFF
--- a/frontend/app/routes/protected/in-person/state-machine-service.ts
+++ b/frontend/app/routes/protected/in-person/state-machine-service.ts
@@ -1,0 +1,114 @@
+import type { ActionFunctionArgs } from 'react-router';
+import { generatePath } from 'react-router';
+
+import type { Actor } from 'xstate';
+import { createActor } from 'xstate';
+
+import { LogFactory } from '~/.server/logging';
+import { AppError } from '~/errors/app-error';
+import { ErrorCodes } from '~/errors/error-codes';
+import { i18nRoutes } from '~/i18n-routes';
+import type { StateName } from '~/routes/protected/in-person/state-machine';
+import { machine } from '~/routes/protected/in-person/state-machine';
+import { getLanguage } from '~/utils/i18n-utils';
+import { getRouteByFile } from '~/utils/route-utils';
+
+type Machine = typeof machine;
+
+const log = LogFactory.getLogger(import.meta.url);
+
+/**
+ * Creates a new in-person application state machine actor and stores it in the session.
+ *
+ * This function retrieves the `tabId` from the request URL, creates a new XState actor
+ * for the in-person application flow, subscribes to its state changes to persist
+ * the snapshot in the session, and starts the actor.
+ *
+ * @throws {Response} 400 response if the `tabId` is missing in the request.
+ */
+export function createMachineActor(session: AppSession, request: Request): Actor<Machine> {
+  const flow = (session.inPersonFlow ??= {}); // ensure session container exists
+
+  const tabId = new URL(request.url).searchParams.get('tid');
+
+  if (!tabId) {
+    log.warn('Could not find tabId in request; returning 400 response.');
+    throw Response.json('The tabId could not be found in the request', { status: 400 });
+  }
+
+  const actor = createActor(machine);
+  actor.subscribe((snapshot) => void (flow[tabId] = snapshot));
+  log.debug('Created new in-person state machine for session [%s] and tabId [%s]', session.id, tabId);
+
+  return actor.start();
+}
+
+/**
+ * Loads an existing in-person application state machine actor from the session.
+ *
+ * This function retrieves the `tabId` from the request URL, attempts to load a
+ * persisted XState actor snapshot from the session, and recreates the actor
+ * with the loaded snapshot.
+ *
+ * @throws {AppError} If the machine snapshot is not found in the session.
+ * @throws {Response} 400 response if the `tabId` is missing in the request.
+ */
+export function loadMachineActor(session: AppSession, request: Request, state: StateName): Actor<Machine> {
+  const flow = (session.inPersonFlow ??= {}); // ensure session container exists
+
+  const tabId = new URL(request.url).searchParams.get('tid');
+
+  if (!tabId) {
+    // XXX :: GjB :: how should we handle this case?
+    //               Showing an error page seems like bad UX.
+    log.warn('Could not find tabId in request; returning 400 response.');
+    throw Response.json('The tabId could not be found in the request', { status: 400 });
+  }
+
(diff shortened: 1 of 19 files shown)


frontend/app/routes/protected/in-person/state-machine-service.ts
sebastien-comeau, line 1: Add `.server` suffix to the filename since it should only be used server-side.

```diff
-frontend/app/routes/protected/in-person/state-machine-service.ts
+frontend/app/routes/protected/in-person/state-machine-service.server.ts
```

frontend/app/routes/protected/in-person/state-machine.ts
sebastien-comeau, line 1: Add `.server` suffix to the filename since it should only be used server-side.

```diff
-frontend/app/routes/protected/in-person/state-machine.ts
+frontend/app/routes/protected/in-person/state-machine.server.ts
```

frontend/app/routes/protected/in-person/state-machine.ts
```diff
@@ -1,194 +1,143 @@
...
+  'primary-docs': 'routes/protected/in-person/primary-docs.tsx',
+  'secondary-docs': 'routes/protected/in-person/secondary-docs.tsx',
+  'name-info': 'routes/protected/in-person/name-info.tsx',
+  'personal-info': 'routes/protected/in-person/personal-info.tsx',
+  'birth-info': 'routes/protected/in-person/birth-info.tsx',
+  'parent-info': 'routes/protected/in-person/parent-info.tsx',
+  'previous-sin-info': 'routes/protected/in-person/previous-sin-info.tsx',
+  'contact-info': 'routes/protected/in-person/contact-info.tsx',
+  'review': 'routes/protected/in-person/review.tsx',
+} satisfies Record<StateName, I18nRouteFile>;
```
sebastien-comeau, line 40: Set routes props as `readonly`

```suggestion
} as const satisfies Record<StateName, I18nRouteFile>;
```

frontend/app/routes/protected/in-person/state-machine-service.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  if (!flow[tabId]) {
+    // XXX :: GjB :: how should we handle this case?
+    //               Showing an error page seems like bad UX.
+    throw new AppError(
+      'The machine snapshot was not found in session', //
+      ErrorCodes.MISSING_SNAPSHOT,
+    );
+  }
```
sebastien-comeau, line 75: You log warnings and you do one of the following options:

1. `throw new Response("Not Found", { status: 404 });`
2. `throw i18nRedirect('routes/protected/index.tsx', request);`

sebastien-comeau (changes requested):

frontend/app/routes/protected/in-person/state-machine-service.ts
gregory-j-baker: Sure.. I was relying on vite to tree-shake it out of the client bundle, but I guess it wouldn't hurt to enforce this with a `.server` qualifier.

frontend/app/routes/protected/in-person/state-machine.ts
gregory-j-baker: Good catch. I didn't have it originally because I wasn't exporting it.. I forgot to add it when exporting.

frontend/app/routes/protected/in-person/state-machine-service.ts
gregory-j-baker: Re: throwing a redirect

I think this is probably the correct approach, but I think we should consider adding some sort of message/notification so the user isn't surprised when it happens.

Imagine the scenario where a user bookmarks a step and then tries to come back to it later on. If we redirect back to the start, I feel like a message would be useful to at least let the user know why the page she bookmarked isn't being rendered.

frontend/app/routes/protected/in-person/state-machine.ts
gregory-j-baker: 07517daeb66a9aa407e7942dde2e3794a2bcb9ac

frontend/app/routes/protected/in-person/state-machine.ts
gregory-j-baker: 07517daeb66a9aa407e7942dde2e3794a2bcb9ac

frontend/app/routes/protected/in-person/state-machine-service.ts
gregory-j-baker: What do you think about what I did in 07517daeb66a9aa407e7942dde2e3794a2bcb9ac?

sebastien-comeau: > Imagine the scenario where a user bookmarks a step and then tries to come back to it later on. If we redirect back to the start, I feel like a message would be useful to at least let the user know why the page she bookmarked isn't being rendered.

That scenario doesn't make sense in an administrative application where things are constantly changing. Additionally, the specific case we're currently working on involves creating a new case, and all routes remain accessible throughout the creation process.

sebastien-comeau (approved):

frontend/app/routes/protected/in-person/state-machine-service.ts
gregory-j-baker: I don't disagree, but when the application does something unexpected (like redirect to `index.tsx` seemingly out of the blue), I think it's at least considerate to notify the user why.

My example was contrived and not really realistic, but I still think we should handle the situation gracefully.

sebastien-comeau: We do by redirecting.

gregory-j-baker: Let me rephrase:

> ...but I still think we should handle the situation gracefully *from the user's perspective*.

PR #192, opened 2025-02-07
(frontend): add document type field on primary identification screen

faiza-jahanzeb:
## Summary

Work item: [#14682](https://dev.azure.com/ESDCCM/FSIR/_workitems/edit/14682)
Add document type field on primary identification screen
User is presented with the Current Status options. Based on the correct options "Canadian citizen born outside Canada" or "Registered Indian born in Canada" in current status, the document type show respective options.
On pressing next button, field is validated. Correct validation for proof of concept is "Canadian citizen born outside Canada"

The document type is validated for null value.

## Types of changes

## Checklist

<details>

  ``` shell
  npm run lint:check
  npm run format:check
  ```
</details>

<details>

  ``` shell
  npm run test
  npm run test:e2e
  ```
</details>

image
image

### DIFF
--- a/frontend/app/routes/protected/person-case/primary-docs.tsx
+++ b/frontend/app/routes/protected/person-case/primary-docs.tsx
@@ -1,9 +1,10 @@
-import { useId } from 'react';
+import { useId, useState } from 'react';
 
 import { data, useFetcher } from 'react-router';
 import type { RouteHandle } from 'react-router';
 
 import { faExclamationCircle, faXmark } from '@fortawesome/free-solid-svg-icons';
+import type { SessionData } from 'express-session';
 import { useTranslation } from 'react-i18next';
 import * as v from 'valibot';
 
@@ -16,9 +17,22 @@ import { FetcherErrorSummary } from '~/components/error-summary';
 import { InputSelect } from '~/components/input-select';
 import { PageTitle } from '~/components/page-title';
 import { Progress } from '~/components/progress';
+import { AppError } from '~/errors/app-error';
+import { ErrorCodes } from '~/errors/error-codes';
 import { getFixedT } from '~/i18n-config.server';
 import { handle as parentHandle } from '~/routes/protected/layout';
 import { getLanguage } from '~/utils/i18n-utils';
+
+type PrimaryDocumentsSessionData = NonNullable<SessionData['inPersonSINCase']['primaryDocuments']>;
+
+/**
+ * Valid current status in Canada for proof of concept
+ */
+const VALID_CURRENT_STATUS = ['canadian-citizen-born-outside-canada'];
+/**
+ * Valid document type for proof of concept
+ */
+const VALID_DOCTYPE = ['certificate-of-canadian-citizenship', 'certificate-of-registration-of-birth-abroad'];
 
 export const handle = {
   i18nNamespace: [...parentHandle.i18nNamespace, 'protected'],
@@ -30,9 +44,7 @@ export async function loader({ context, request }: Route.LoaderArgs) {
 
   return {
     documentTitle: t('protected:primary-identity-document.page-title'),
-    defaultFormValues: {
-      currentStatusInCanada: context.session.inPersonSINCase?.currentStatusInCanada,
-    },
+    defaultFormValues: context.session.inPersonSINCase?.primaryDocuments,
   };
 }
 
@@ -46,37 +58,56 @@ export async function action({ context, request }: Route.ActionArgs) {
   const t = await getFixedT(request, handle.i18nNamespace);
 
   const formData = await request.formData();
+  const action = formData.get('action');
 
-  if (formData.get('action') === 'back') {
-    throw i18nRedirect('routes/protected/person-case/privacy-statement.tsx', request); //TODO: change it to redirect to file="routes/protected/person-case/request-details.tsx"
+  switch (action) {
+    case 'back': {
+      throw i18nRedirect('routes/protected/person-case/request-details.tsx', request);
+    }
+
+    case 'next': {
+      const schema = v.pipe(
+        v.object({
+          currentStatusInCanada: v.pipe(
+            v.string(t('protected:primary-identity-document.current-status-in-canada.required')),
+            v.trim(),
+            v.nonEmpty(t('protected:primary-identity-document.current-status-in-canada.required')),
+            v.picklist(VALID_CURRENT_STATUS, t('protected:primary-identity-document.current-status-in-canada.invalid')),
+          ),
+          documentType: v.string(),
+        }),
+        // Perform an additional check on 'documentType' only when currentStatusInCanada is valid.
+        // Otherwise, it incorrectly triggers an error even when 'documentType' is not visible.
+        v.forward(
+          v.partialCheck(
+            [['currentStatusInCanada'], ['documentType']],
+            (input) => VALID_DOCTYPE.includes(input.documentType),
+            t('protected:primary-identity-document.document-type.required'),
+          ),
+          ['documentType'],
+        ),
+      ) satisfies v.GenericSchema<PrimaryDocumentsSessionData>;
+
+      const input = {
+        currentStatusInCanada: String(formData.get('currentStatusInCanada')),
+        documentType: String(formData.get('documentType')),
+      } satisfies Partial<v.InferInput<typeof schema>>;
+
+      const parseResult = v.safeParse(schema, input, { lang });
+
+      if (!parseResult.success) {
+        return data({ errors: v.flatten<typeof schema>(parseResult.issues).nested }, { status: 400 });
+      }
+
+      context.session.inPersonSINCase ??= {};
+      context.session.inPersonSINCase.primaryDocuments = parseResult.output;
+
+      throw i18nRedirect('routes/protected/request.tsx', request); //TODO: change it to redirect to file="routes/protected/person-case/secondary-docs.tsx"
+    }
+    default: {
+      throw new AppError(`Unrecognized action: ${action}`, ErrorCodes.UNRECOGNIZED_ACTION);
+    }
   }
-
-  // submit action
-  const schema = v.object({
-    currentStatusInCanada: v.pipe(
-      v.string(t('protected:primary-identity-document.current-status-in-canada.required')),
-      v.trim(),
-      v.nonEmpty(t('protected:primary-identity-document.current-status-in-canada.required')),
-      v.literal(
-        'canadian-citizen-born-outside-canada',
-        t('protected:primary-identity-document.current-status-in-canada.invalid'),
-      ),
-    ),
-  });
-
-  const input = { currentStatusInCanada: formData.get('currentStatusInCanada') as string };
-  const parsedDataResult = v.safeParse(schema, input, { lang });
-
-  if (!parsedDataResult.success) {
-    return data({ errors: v.flatten<typeof schema>(parsedDataResult.issues).nested }, { status: 400 });
-  }
-
-  context.session.inPersonSINCase = {
-    ...(context.session.inPersonSINCase ?? {}),
-    ...input,
-  };
-
-  throw i18nRedirect('routes/protected/request.tsx', request); //TODO: change it to redirect to file="routes/protected/person-case/secondary-docs.tsx"
 }
 
 export default function PrimaryDocs({ loaderData, actionData, params }: Route.ComponentProps) {
@@ -87,41 +118,11 @@ export default function PrimaryDocs({ loaderData, actionData, params }: Route.Co
   const isSubmitting = fetcher.state !== 'idle';
   const errors = fetcher.data?.errors;
 
-  const dummyOption: { label: string; value: string } = {
-    label: t('protected:primary-identity-document.please-select'),
-    value: '',
+  const [currentStatus, setCurrentStatus] = useState(loaderData.defaultFormValues?.currentStatusInCanada);
+
+  const handleCurrentStatusChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setCurrentStatus(event.target.value);
   };
-  const currentStatusInCanadaOptions: { label: string; value: string }[] = [
-    dummyOption,
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.canadian-citizen-born-in-canada'),
-      value: 'canadian-citizen-born-in-canada',
-    },
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.canadian-citizen-born-outside-canada'),
-      value: 'canadian-citizen-born-outside-canada',
-    },
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.registered-indian-born-in-canada'),
-      value: 'registered-indian-born-in-canada',
-    },
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.registered-indian-born-outside-canada'),
-      value: 'registered-indian-born-outside-canada',
-    },
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.permanent-resident'),
-      value: 'permanent-resident',
-    },
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.temporary-resident'),
-      value: 'temporary-resident',
-    },
-    {
-      label: t('protected:primary-identity-document.current-status-in-canada.options.no-legal-status-in-canada'),
-      value: 'no-legal-status-in-canada',
-    },
-  ];
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+          v.picklist(VALID_CURRENT_STATUS, t('protected:primary-identity-document.current-status-in-canada.invalid')),
+        ),
+      });
+      input.currentStatusInCanada = formData.get('currentStatusInCanada') as string;
+      const currentStatusResult = v.safeParse(currentStatusSchema, input, { lang });
 
-  if (!parsedDataResult.success) {
-    return data({ errors: v.flatten<typeof schema>(parsedDataResult.issues).nested }, { status: 400 });
-  }
+      let documentTypeResult = null;
```
gregory-j-baker, line 76: This can be moved to where it is initialized (the `v.safeParse()` line).

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+            {
+              errors: {
+                ...(currentStatusResult.issues ? v.flatten<typeof currentStatusSchema>(currentStatusResult.issues).nested : {}),
+                ...v.flatten<typeof documentTypeSchema>(documentTypeResult.issues).nested,
+              },
+            },
+            { status: 400 },
+          );
+        }
+      } else if (!currentStatusResult.success) {
```
gregory-j-baker, line 100: Bit of a nitpick, but this if/else could be simplified a bit by moving the `else if` up and returning early.. I think this would improve readability.

ex:

``` typescript
if (!currentStatusResult.success) {
  return data({ errors: v.flatten<typeof currentStatusSchema>(currentStatusResult.issues).nested }, { status: 400 });
}

// rest of function without the need for "else"
```

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
 
 export default function PrimaryDocs({ loaderData, actionData, params }: Route.ComponentProps) {
   const { t } = useTranslation(handle.i18nNamespace);
 
   const fetcher = useFetcher<Info['actionData']>();
   const isSubmitting = fetcher.state !== 'idle';
-  const errors = fetcher.data?.errors;
+  const errors: { currentStatusInCanada?: string[]; documentType?: string[] } | undefined = fetcher.data?.errors ?? undefined;
+
+  const [step, setStep] = useState(1);
```
gregory-j-baker, line 125: Try to avoid using magic numbers.. replace `1` with something more descriptive and meaningful.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+          <DocumentTypeRegisteredIndian
+            defaultValue={loaderData.defaultFormValues.documentType}
+            errorMessage={errors?.documentType?.[0]}
+          />,
+        );
+        break;
+      default:
+        setSecondaryInput(null); // Clear secondary input if no match
+    }
+  }, [currentStatus, errors, loaderData.defaultFormValues.documentType]);
```
gregory-j-baker, line 159: I think a better approach here would be to use a flag to indicate what the secondary input should be, and then render it conditionally based on that flag. Maybe I'm wrong.. it just seems a bit unconventional to store a rendered component in another component's state.

Also, you shouldn't need to use an effect for this. In fact, `useEffect()` in React Router v7 should rarely be needed for anything.

From <https://react.dev/reference/react/useEffect>:

image

You aren't synchronizing with an external system here.

(will ping @sebastien-comeau here since he has more expertise around this stuff)

gregory-j-baker: To elaborate on the "you shouldn't need effects in react router" comment.. when using react router, you typically synchronize with your external systems in your actions and loaders.

There are cases where an effect might be needed (like when making calls to a browser API), but in pretty much every other situation, you can use actions/loaders and conditional rendering.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+        </Button>
+        <Button id="refer-button" endIcon={faExclamationCircle} variant="link">
+          {t('protected:person-case.refer-button')}
+        </Button>
+      </div>
+      <Progress className="mt-8" label="" value={30} />
+      <PageTitle subTitle={t('protected:in-person.title')}>{t('protected:primary-identity-document.page-title')}</PageTitle>
+
+      <fetcher.Form method="post" noValidate>
+        <input type="hidden" name="step" value={step.toString()} />
```
gregory-j-baker, line 175: I don't think a hidden field is necessary here if you are using `useState()` to handle rendering of the secondary input. You should just be able to call `setState('my-value')` and then render the second input by examining that value.

You may not even need to use `useState()` if you render conditionally based on the value of your `<CurrentStatusInCanada>` component.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+  );
+}
+
+interface CurrentStatusInCanadaProps {
+  defaultValue?: string;
+  errorMessage?: string;
+  onChange?: React.ChangeEventHandler<HTMLSelectElement>;
+}
+
+function CurrentStatusInCanada({ defaultValue, errorMessage, onChange }: CurrentStatusInCanadaProps) {
```
gregory-j-baker, line 213: (nitpick)

```suggestion
function CurrentStatusInCanada({ defaultValue, errorMessage, onChange }: CurrentStatusInCanadaProps): JSX.Element {
```

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -122,40 +251,96 @@ export default function PrimaryDocs({ loaderData, actionData, params }: Route.Co
...
-          </Button>
-          <Button name="action" value="next" variant="primary" id="continue-button" disabled={isSubmitting}>
-            {t('protected:person-case.next')}
-          </Button>
-        </div>
-      </fetcher.Form>
+function DocumentType({ defaultValue, errorMessage }: DocumentTypeProps) {
+  const { t } = useTranslation(handle.i18nNamespace);
+
+  const dummyOption: { label: string; value: string } = {
```
gregory-j-baker, line 276: I'm pretty sure the type is not required here, since it will be inferred by the values.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -44,45 +52,166 @@ export async function action({ context, request }: Route.ActionArgs) {
...
 
-  const input = { currentStatusInCanada: formData.get('currentStatusInCanada') as string };
-  const parsedDataResult = v.safeParse(schema, input, { lang });
+      const currentStatusSchema = v.object({
+        currentStatusInCanada: v.pipe(
+          v.string(t('protected:primary-identity-document.current-status-in-canada.required')),
+          v.trim(),
+          v.nonEmpty(t('protected:primary-identity-document.current-status-in-canada.required')),
+          v.picklist(VALID_CURRENT_STATUS, t('protected:primary-identity-document.current-status-in-canada.invalid')),
+        ),
```
sebastien-comeau, line 71: Validate with `picklist`.

```suggestion
        currentStatusInCanada: v.picklist(VALID_CURRENT_STATUS, t('protected:primary-identity-document.current-status-in-canada.required')),
```

sebastien-comeau (changes requested):
I'd need to experiment with it, but I think using a single validation and returning all errors to the component, even if the entire form isn't displayed, could work. The "step" logic might not be necessary.

frontend/app/routes/protected/person-case/primary-docs.tsx
faiza-jahanzeb: removed useEffect()

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -85,43 +123,13 @@ export default function PrimaryDocs({ loaderData, actionData, params }: Route.Co
   const fetcherKey = useId();
   const fetcher = useFetcher<Info['actionData']>({ key: fetcherKey });
   const isSubmitting = fetcher.state !== 'idle';
-  const errors = fetcher.data?.errors;
+  const errors: { currentStatusInCanada?: string[]; documentType?: string[] } | undefined = fetcher.data?.errors ?? undefined;
```
gregory-j-baker, line 126: The `?? undefined` is unnecessary here.

gregory-j-baker (dismissed):

frontend/app/routes/protected/person-case/primary-docs.tsx
gregory-j-baker: Will resolve since this isn't a blocker

frontend/app/routes/protected/person-case/primary-docs.tsx
sebastien-comeau: Why `errors` variable is typed with `: { currentStatusInCanada?: string[]; documentType?: string[] } | undefined`? This should work.

```tsx
const errors = fetcher.data?.errors;
```

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -46,37 +57,64 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+              documentType: v.optional(v.string()),
+            }),
+            v.object({
+              documentTypeAvailable: v.literal(true),
+              documentType: v.pipe(v.picklist(VALID_DOCTYPE, t('protected:primary-identity-document.document-type.required'))),
+            }),
+          ],
+          t('protected:primary-identity-document.document-type.required'),
+        ),
+      ]) satisfies v.GenericSchema<PrimaryDocumentsSessionData>;
```
sebastien-comeau, line 92: Validation schema can be simplified like this. `currentStatusInCanada` and `documentType` are both required field and must be a valid string value from the `picklist`. There is nothing bad to validate `documentType` even though it's not displayed on the screen. The error for that field will not be displayed.

```suggestion
      const schema = v.object({
          currentStatusInCanada: v.picklist(VALID_CURRENT_STATUS, t('protected:primary-identity-document.current-status-in-canada.required')),
            documentType: v.picklist(VALID_DOCTYPE, t('protected:primary-identity-document.document-type.required')),
            }) satisfies v.GenericSchema<PrimaryDocumentsSessionData>;
```

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -46,37 +57,64 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+      if (!parseResult.success) {
+        return data({ errors: v.flatten<typeof schema>(parseResult.issues).nested }, { status: 400 });
+      }
 
-  throw i18nRedirect('routes/protected/request.tsx', request); //TODO: change it to redirect to file="routes/protected/person-case/secondary-docs.tsx"
+      context.session.inPersonSINCase ??= {};
+      context.session.inPersonSINCase.primaryDocuments = {
+        ...parseResult.output,
+        documentType: parseResult.output.documentType ?? '',
+      };
```
sebastien-comeau, line 110: Entire validated validation should be stored as is.

```suggestion
      context.session.inPersonSINCase.primaryDocuments = parseResult.output;
```

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -135,17 +143,23 @@ export default function PrimaryDocs({ loaderData, actionData, params }: Route.Co
...
-            errorMessage={errors?.currentStatusInCanada?.at(0)}
-            defaultValue={loaderData.defaultFormValues.currentStatusInCanada}
-            required
-            options={currentStatusInCanadaOptions}
-            label={t('protected:primary-identity-document.current-status-in-canada.title')}
-          />
+          <div className="space-y-6">
+            <CurrentStatusInCanada
+              defaultValue={loaderData.defaultFormValues?.currentStatusInCanada}
+              errorMessage={errors?.currentStatusInCanada?.[0]}
```
sebastien-comeau, line 152: Use [at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at).

```suggestion
              errorMessage={errors?.currentStatusInCanada?.at(0)}
```

sebastien-comeau (changes requested):

frontend/app/routes/protected/person-case/primary-docs.tsx
faiza-jahanzeb: without at() works... errorMessage={errors?.currentStatusInCanada?.at(0)} cause syntax error

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -1,7 +1,7 @@
-import { useId } from 'react';
+import { useId, useState } from 'react';
 
 import { data, useFetcher } from 'react-router';
-import type { RouteHandle } from 'react-router';
+import type { RouteHandle, SessionData } from 'react-router';
```
sebastien-comeau, line 4: Import `SessionData` from `express-session`. We are augmenting `SessionData` from `express-session` and not `react-router`. Data type of `PrimaryDocumentsSessionData` is `any` at the moment.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -16,10 +16,23 @@ import { FetcherErrorSummary } from '~/components/error-summary';
...
 import { getFixedT } from '~/i18n-config.server';
 import { handle as parentHandle } from '~/routes/protected/layout';
 import { getLanguage } from '~/utils/i18n-utils';
 
+type PrimaryDocumentsSessionData = NonNullable<SessionData['inPersonSINCase']['primaryDocuments']>;
+
+/**
+ * Valid current status in Canada for proof of concept
+ */
+const VALID_CURRENT_STATUS = ['canadian-citizen-born-outside-canada'] as const;
```
sebastien-comeau, line 30: Remove `as const` since those values doesn't need to be string literals and it causes issue while build the validation schema.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -16,10 +16,23 @@ import { FetcherErrorSummary } from '~/components/error-summary';
...
+/**
+ * Valid document type for proof of concept
+ */
+const VALID_DOCTYPE = ['certificate-of-canadian-citizenship', 'certificate-of-registration-of-birth-abroad'] as const;
```
sebastien-comeau, line 34: Remove `as const` since those values doesn't need to be string literals and it causes issue while build the validation schema.

sebastien-comeau (changes requested):

frontend/app/routes/protected/person-case/primary-docs.tsx
sebastien-comeau: I tested it locally, and since 'documentType' might not be visible, returning an error when it's hidden causes the error to immediately appear when the field becomes visible. It feels odd. Here's an improved validation approach:

```ts
const schema = v.pipe(
  v.object({
    currentStatusInCanada: v.picklist(
      VALID_CURRENT_STATUS,
      t('protected:primary-identity-document.current-status-in-canada.required'),
    ),
    documentType: v.string(),
  }),
  // Perform an additional check on 'documentType' only when currentStatusInCanada is valid.
  // Otherwise, it incorrectly triggers an error even when 'documentType' is not visible.
  v.forward(
    v.partialCheck(
      [['currentStatusInCanada'], ['documentType']],
      (input) => VALID_DOCTYPE.includes(input.documentType),
      t('protected:primary-identity-document.document-type.required'),
    ),
    ['documentType'],
  ),
) satisfies v.GenericSchema<PrimaryDocumentsSessionData>;

const input = {
  currentStatusInCanada: String(formData.get('currentStatusInCanada')),
  documentType: String(formData.get('documentType')),
} satisfies Partial<v.InferInput<typeof schema>>;
```

faiza-jahanzeb: updated

sebastien-comeau (approved):

PR #222, opened 2025-02-17
(frontend): primary document fields added

faiza-jahanzeb:
## Summary

work item # [14682](https://dev.azure.com/ESDCCM/FSIR/_workitems/edit/14682)
Added more fields to the Primary document screen, as displayed in figma.

On selecting these values in these fields:
Current status in Canada:  "Canadian citizen born outside Canada" (rest of the options are disabled)
Document type: "Certificate of Canadian Citizenship" (rest of the options are disabled)

More fields are displayed (and validated after input):
Registration number (8 digit number)
Client number (10 digit number)
Given name (non digit alphabets lass than 100 characters)
Last name (non digit alphabets lass than 100 characters)
Date of birth (should be a date in past or today)
Gender (options: female, male, other)
Citizenship date (should be a date)
Upload primary identification document. (File)

## Types of changes

## Checklist

## Screenshots 

image

### DIFF
--- a/frontend/app/routes/protected/person-case/primary-docs.tsx
+++ b/frontend/app/routes/protected/person-case/primary-docs.tsx
@@ -1,3 +1,4 @@
+import type { JSX } from 'react';
 import { useId, useState } from 'react';
 
 import type { RouteHandle } from 'react-router';
@@ -10,10 +11,15 @@ import * as v from 'valibot';
 
 import type { Info, Route } from './+types/primary-docs';
 
+import { serverEnvironment } from '~/.server/environment';
 import { requireAuth } from '~/.server/utils/auth-utils';
 import { i18nRedirect } from '~/.server/utils/route-utils';
 import { Button } from '~/components/button';
+import { DatePickerField } from '~/components/date-picker-field';
 import { FetcherErrorSummary } from '~/components/error-summary';
+import { InputField } from '~/components/input-field';
+import { InputFile } from '~/components/input-file';
+import { InputRadios } from '~/components/input-radios';
 import { InputSelect } from '~/components/input-select';
 import { PageTitle } from '~/components/page-title';
 import { Progress } from '~/components/progress';
@@ -21,12 +27,22 @@ import { AppError } from '~/errors/app-error';
 import { ErrorCodes } from '~/errors/error-codes';
 import { getTranslation } from '~/i18n-config.server';
 import { handle as parentHandle } from '~/routes/protected/layout';
+import { getStartOfDayInTimezone, isDateInPastOrTodayInTimeZone, isValidDateString, toISODateString } from '~/utils/date-utils';
+import { REGEX_PATTERNS } from '~/utils/regex-utils';
 
 type PrimaryDocumentsSessionData = NonNullable<SessionData['inPersonSINCase']['primaryDocuments']>;
 
 const VALID_CURRENT_STATUS = ['canadian-citizen-born-outside-canada'];
+/**
+ * Valid document type for proof of concept
+ */
+const VALID_DOCTYPES = ['certificate-of-canadian-citizenship'];
+/**
+ * Valid gender for primary identification document
+ */
+const VALID_GENDERS = ['female', 'male', 'other'];
 
-const VALID_DOCTYPE = ['certificate-of-canadian-citizenship', 'certificate-of-registration-of-birth-abroad'];
+const timezone = serverEnvironment.BASE_TIMEZONE;
 
 export const handle = {
   i18nNamespace: [...parentHandle.i18nNamespace, 'protected'],
@@ -52,6 +68,10 @@ export async function action({ context, request }: Route.ActionArgs) {
   const { lang, t } = await getTranslation(request, handle.i18nNamespace);
   const formData = await request.formData();
   const action = formData.get('action');
+  const nameMaxLength = 100;
+  const registrationNumberLength = 8;
+  const clientNumberLength = 10;
+  const maxImageSizeBits = 1024 * 1024 * 15; //Max image size is 15 MB
 
   switch (action) {
     case 'back': {
@@ -59,7 +79,9 @@ export async function action({ context, request }: Route.ActionArgs) {
     }
 
     case 'next': {
-      const schema = v.pipe(
+      const maxAllowedDate = getStartOfDayInTimezone(timezone);
+
+      const schema = v.intersect([
         v.object({
           currentStatusInCanada: v.pipe(
             v.string(t('protected:primary-identity-document.current-status-in-canada.required')),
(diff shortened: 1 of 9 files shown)


frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -59,44 +69,176 @@ export async function action({ context, request }: Route.ActionArgs) {
...
-      const schema = v.pipe(
+      const dateFormatEST = new Intl.DateTimeFormat('en-CA', {
+        timeZone: 'America/Toronto',
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit',
+      });
+
+      //Normalized date at current day midnight EST
+      const currentDate = new Date(`${dateFormatEST.format(new Date())}T00:00:00`);
```
gregory-j-baker, line 91: In Javascript, the current date is internally represented as UTC, so there's no reason to normalize it.

What you *do* need to normalize is what you are comparing it to. So if you are checking that `now < expiryDate`, it's `expiryDate` that you have to normalize.

You want to validate that right now, it's before midnight of `expiryDate` in the eastern timezone.

It's kind of hard to explain, so I asked Google Gemini to ELI5:

> You define the end time (expiry date) relative to EST. JavaScript converts this (internally) to UTC.  `Date.now()` gives you the current UTC time.  JavaScript compares these two UTC values.  You don't normalize `Date.now()` because it's already an instant in time (UTC).

gregory-j-baker: To summarize and be clear:

Always use `Date.now()` (or some equivalent) to get the current instant in time. This value is the precise number of milliseconds since the unix epoch.. perhaps counter-intuitively: **the timezone that you are in is actually completely irrelevant** (presuming the system clock is correct).

When you specify the expiry date, set the initial value to `yyyy-mm-ddT00:00:00` **of the timezone you're interested in**.

gregory-j-baker: As I read through your PR, I realize that in this case you actually are using this value correctly. Sorry.

I will leave the comment since it is still valid and contains useful information.

One thing I would suggest, though.. is to maybe rename this variable to something a little more descriptive. Perhaps `maxDate` or `maxAllowedDate`. This would prevent future morons like me misinterpreting its use like I did here.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -59,44 +69,176 @@ export async function action({ context, request }: Route.ActionArgs) {
...
-            t('protected:primary-identity-document.document-type.required'),
-          ),
-          ['documentType'],
+        v.variant(
+          'documentType',
+          [
+            v.object({
+              documentType: v.picklist(VALID_DOCTYPE, t('protected:primary-identity-document.document-type.invalid')),
+
+              // primary document fields
```
gregory-j-baker, line 108: This comment adds nothing useful.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -59,44 +69,176 @@ export async function action({ context, request }: Route.ActionArgs) {
...
         currentStatusInCanada: String(formData.get('currentStatusInCanada')),
         documentType: String(formData.get('documentType')),
+        registrationNumber: String(formData.get('registrationNumber')),
+        clientNumber: String(formData.get('clientNumber')),
+        givenName: String(formData.get('givenName')),
+        lastName: String(formData.get('lastName')),
+        dateOfBirthYear: dateOfBirthYear,
+        dateOfBirthMonth: dateOfBirthMonth,
+        dateOfBirthDay: dateOfBirthDay,
+        dateOfBirth: new Date(dateOfBirthYear, dateOfBirthMonth - 1, dateOfBirthDay), // The month as a number between 0 and 11 (January to December)
```
gregory-j-baker, line 229: This should be represented as a string, in my opinion.. Or, if you really must store it as a date, then it should be normalized to eastern time.. otherwise it will be the birthdate at `T00:00:00Z` which isn't correct.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -59,44 +69,176 @@ export async function action({ context, request }: Route.ActionArgs) {
...
+        gender: String(formData.get('gender')),
+        citizenshipDateYear: citizenshipDateYear,
+        citizenshipDateMonth: citizenshipDateMonth,
+        citizenshipDateDay: citizenshipDateDay,
+        citizenshipDate: new Date(citizenshipDateYear, citizenshipDateMonth - 1, citizenshipDateDay), // The month as a number between 0 and 11 (January to December)
```
gregory-j-baker, line 234: This should be represented as a string, in my opinion.. Or, if you really must store it as a date, then it should be normalized to eastern time.. otherwise it will be the birthdate at `T00:00:00Z` which isn't correct.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -268,8 +427,132 @@ function DocumentType({ currentStatus, defaultValue, errorMessage }: DocumentTyp
...
+    lastName: string;
+    registrationNumber: string;
+  };
+  documentType?: string;
+  errors?: Record<string, [string, ...string[]] | undefined>;
+}
+
+function PrimaryDocsFields({ currentStatus, defaultValues, errors, documentType }: PrimaryDocsFieldsProps): JSX.Element {
+  const { t } = useTranslation(handle.i18nNamespace);
+  const GENDER = ['female', 'male', 'other'] as const;
```
gregory-j-baker, line 454: This should be plural.

**Also, nitpick:** there's no reason to make it all caps. We don't follow this convention elsewhere in the app.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -268,8 +427,132 @@ function DocumentType({ currentStatus, defaultValue, errorMessage }: DocumentTyp
...
         />
       )}
     </>
   );
 }
+
+interface PrimaryDocsFieldsProps {
+  currentStatus?: string;
+  defaultValues?: {
+    citizenshipDate: Date;
```
gregory-j-baker, line 440: This should be a string; there's no good reason for it to be a javascript `Date`.

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -268,8 +427,132 @@ function DocumentType({ currentStatus, defaultValue, errorMessage }: DocumentTyp
...
+    clientNumber: string;
+    dateOfBirth: Date;
```
gregory-j-baker, line 442: This should be a string; there's no good reason for it to be a javascript `Date`.

faiza-jahanzeb: updated

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +51,73 @@ export function getLocalizedMonths(
...
+ * Parses a date (year, month, day) and returns string in the format "YYYY-MM-DD".
+ * @param year - The year as number.
+ * @param month - The month as number.
+ * @param day - The day as number.
+ * @returns A string in the format "YYYY-MM-DD" (year, month, day).
+ */
+export function extractDatePartsAsString(year: number, month: number, day: number): string {
+  const dateString = String(padWithZero(year, 4)) + '-' + String(padWithZero(month, 2)) + '-' + String(padWithZero(day, 2));
+  return dateString;
+}
```
gregory-j-baker, line 123: I hate to be nitpicky, but I feel like the name of this function doesn't really describe what it's doing. I think a better name would be something like `toISODateString()`.

Also, since `padWithZero()` already returns a string, you don't need to call `String()` to cast it.

I would write this function like this:

``` ts
export function toISODateString(year: number, month: number, day: number): string {
￼  return `${padWithZero(year, 4)}-${padWithZero(month, 2)}-${padWithZero(day, 2)}`;
￼}
```

(preferring string templates over `string + string` syntax)

faiza-jahanzeb: updated

gregory-j-baker (dismissed):

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +51,76 @@ export function getLocalizedMonths(
...
+ * console.log(isValidDateString(invalidDate)); // false
+ * ```
+ */
+export const isValidDateString = (dateString: string) => {
+  const dateRegex = /^\d{4}-\d{2}-\d{2}$/;
+  if (!dateRegex.test(dateString)) {
+    return false;
+  }
+  const date = new Date(dateString);
+  return !isNaN(date.getTime());
```
sebastien-comeau, line 76: Can be simplified with `date-fns` functions: [isValid](https://date-fns.org/v4.1.0/docs/isValid) and [parseISO](https://date-fns.org/v4.1.0/docs/parseISO)

```suggestion
  return isValid(parseISO(dateString));
```

frontend/app/routes/protected/person-case/primary-docs.tsx
```diff
@@ -59,44 +70,192 @@ export async function action({ context, request }: Route.ActionArgs) {
...
-      const schema = v.pipe(
+      const dateFormatEST = new Intl.DateTimeFormat('en-CA', {
+        timeZone: 'America/Toronto',
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit',
+      });
+
+      //Normalized date at current day midnight EST
+      const maxAllowedDate = new Date(`${dateFormatEST.format(new Date())}T00:00:00`);
```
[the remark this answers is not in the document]
gregory-j-baker:
``` ts
function getStartOfDayInTimezone(timezone: string, date?: number | string | Date): Date {
  return new Date(startOfDay(TZDate.tz(timezone, new Date(date ?? Date.now()))).toISOString());
}

const maxAllowedDate = getStartOfDayInTimezone('Canada/Eastern')
```

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +51,76 @@ export function getLocalizedMonths(
...
+    timeZone: 'America/Toronto',
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+  });
+
+  const maxAllowedDate = new Date(`${dateFormatEST.format(new Date())}T00:00:00`);
+
+  if (inputDate <= maxAllowedDate) return true;
+  return false;
```
sebastien-comeau, line 114: Use [isPast](https://date-fns.org/v4.1.0/docs/isPast) and [isToday](https://date-fns.org/v4.1.0/docs/isToday) from `date-fns`.

```suggestion
  const inputDate = parseISO(dateString);
  return isPast(inputDate) || isToday(inputDate);
```

gregory-j-baker: We can't use `isPast()` because we have to normalize for a specific timezone. Instead, you have to do something like this:

``` ts
isBefore(inputDate, getStartOfDayInTimezone('Canada/Eastern'))
```

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+ * @example
+ * ```javascript
+ * const validDateInPaste = "2023-05-20";
+ * const invalidDateInFuture = "6078-05-20";
+ *
+ * console.log(isDateInPastOrToday(validDateInPaste)); // true
+ * console.log(isDateInPastOrToday(invalidDateInFuture)); // false
+ * ```
+ */
+export function isDateInPastOrToday(dateString: string): boolean {
```
gregory-j-baker, line 96: I think this function should be broken into smaller ones.

``` ts
function isPastInTimezone(timezone: string, date: number | string | Date) { .. }

function isTodayInTimezone(timezone: string, date: number | string | Date) { .. }

function isPastOrTodayInTimezone(timezone: string, date: number | string | Date) {
  return isPastInTimezone(..) || isTodayInTimezone(..)
}
```

The timezone should be an argument because we shouldn't hardcode the timezone in case we need to change it at runtime (via an environment variable).

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+ * const validDate = "2023-05-20";
+ * const invalidDate = "20-05-2023";
+ *
+ * console.log(isValidDateString(validDate)); // true
+ * console.log(isValidDateString(invalidDate)); // false
+ * ```
+ */
+export const isValidDateString = (dateString: string) => {
+  return isValid(parseISO(dateString));
+};
```
gregory-j-baker, line 75:
```suggestion
export function isValidDateString(dateString: string) {
  return isValid(parseISO(dateString));
}
```

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+ * console.log(isDateInPastOrToday(invalidDateInFuture)); // false
+ * ```
+ */
+export function isDateInPastOrToday(dateString: string): boolean {
+  if (!isValidDateString(dateString)) {
+    throw new Error('Invalid date format. Expected YYYY-MM-DD.');
+  }
+  const inputDate = new Date(dateString);
+
+  const maxAllowedDate = getStartOfDayInTimezone('Canada/Eastern');
```
gregory-j-baker, line 102: This shouldn't be hardcoded. It should be configured via environment variables and then passed in as a function parameter.

If doing this every time you need to check a date is too cumbersome, we can create a service class that will look up the configured timezone and then make the call to this function using that.

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+ * const validDateInPaste = "2023-05-20";
+ * const invalidDateInFuture = "6078-05-20";
+ *
+ * console.log(isDateInPastOrToday(validDateInPaste)); // true
+ * console.log(isDateInPastOrToday(invalidDateInFuture)); // false
+ * ```
+ */
+export function isDateInPastOrToday(dateString: string): boolean {
+  if (!isValidDateString(dateString)) {
+    throw new Error('Invalid date format. Expected YYYY-MM-DD.');
```
gregory-j-baker, line 98: Consider throwing an `AppError` here so you can attach an error code.

When we throw AppErrors, the application will automatically catch them and display the error code to the user. The idea is that the user can give this error code to a help desk agent, who will add it to a ticket. 

This will make troubleshooting what went wrong much easier for us.

gregory-j-baker: Alternatively, does it make sense to have this function simply return false if the passed-in date is malformed? Maybe not, but it's something worth considering.

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+  }
+  const inputDate = new Date(dateString);
+
+  const maxAllowedDate = getStartOfDayInTimezone('Canada/Eastern');
+
+  if (inputDate <= maxAllowedDate) return true;
+  return false;
```
gregory-j-baker, line 105:
```suggestion
  return inputDate <= maxAllowedDate;
```

gregory-j-baker: Or..

```suggestion
  return isBefore(inputDate, maxAllowedDate); // from date-fns
```

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+}
+
+/**
+ * Parses a time zone or a date (number or string or Date) and returns a date in the UTC (00:00:00.000Z) format.
+ * @param timeZone - The time zone as string.
+ * @returns A date set to the start of the day in UTC (00:00:00.000Z).
+ */
+export function getStartOfDayInTimezone(timezone: string, date?: number | string | Date): Date {
+  return new Date(startOfDay(TZDate.tz(timezone, new Date(date ?? Date.now()))));
+}
```
gregory-j-baker, line 115: The JSDoc doesn't really describe what the function is doing. Here's an updated version (generated by Google Gemini):

```suggestion
/**
 * Gets the start of the day in a specific timezone.
 *
 * This function calculates the start of the day (midnight) in the given timezone,
 * based on an optional date. If no date is provided, it uses the current date.
 *
 * @param timezone - The IANA time zone string (e.g., 'Canada/Eastern').
 * @param date - Optional. A date, timestamp (milliseconds since epoch), or a string parseable by `Date.parse()`.
 * @returns A Date object representing the start of the day (midnight) in the specified timezone.
 */
export function getStartOfDayInTimezone(timezone: string, date?: number | string | Date): Date {
  return new Date(startOfDay(TZDate.tz(timezone, new Date(date ?? Date.now()))));
}
```

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+/**
+ * Parses a date (year, month, day) and returns string in the format "YYYY-MM-DD".
+ * @param year - The year as number.
+ * @param month - The month as number.
+ * @param day - The day as number.
+ * @returns A string in the format "YYYY-MM-DD" (year, month, day).
+ */
+export function toISODateString(year: number, month: number, day: number): string {
+  return `${padWithZero(year, 4)}-${padWithZero(month, 2)}-${padWithZero(day, 2)}`;
+}
```
gregory-j-baker, line 126:
```suggestion
/**
 * Formats a date into an ISO8601 date string (YYYY-MM-DD).
 *
 * This function takes the year, month, and day as numerical inputs and
 * returns a string representing the date in the "YYYY-MM-DD" format.
 *
 * @param year - The year (e.g., 2023).
 * @param month - The month (1 for January, 12 for December).
 * @param day - The day of the month.
 * @returns An ISO 8601 date string in the format "YYYY-MM-DD".
 */
export function toISODateString(year: number, month: number, day: number): string {
  return `${padWithZero(year, 4)}-${padWithZero(month, 2)}-${padWithZero(day, 2)}`;
}
```

Also.. I'd like to note that if you give this function something silly like `toISODateString(123, 3.1415, -Infinity)`, it will happily spit out `0123-3.1415--Infinity`. 😏

frontend/app/utils/date-utils.ts
```diff
@@ -51,3 +54,73 @@ export function getLocalizedMonths(
...
+
+/**
+ * Validates whether a given string is in past or today.
+ *
+ * @param date - The date string to be validated.
+ * @returns Returns `true` if the string is a valid date in the format YYYY-MM-DD and the date is in past or today's date, otherwise `false`.
+ *
+ * This function first validates the input date string using the `isValidDateString` function.
+ * If the date string is invalid, an error is thrown. Otherwise, it compare date
+ * object with the date string set to the start of the day in UTC (00:00:00.000Z).
```
gregory-j-baker, line 85: "*it compares date object with the date string set to the start of the day in UTC*" is inaccurate here. It's actually comparing to the start of the day in the target timezone.

frontend/app/utils/date-utils.ts
faiza-jahanzeb: we are getting the date (year as number, month as number and day s number) returned by the date picker field from the form. This value need to be converted to string before all the validations are done on that string. Maybe we need to change the name of this function??

frontend/app/utils/date-utils.ts
faiza-jahanzeb: I applied this suggestion...

frontend/app/utils/date-utils.ts
gregory-j-baker: I think the function name is fine.. it's generating an ISO date string, just as the name suggests.

I think some guards should be added that prevent wonky inputs, though. It should probably throw if the year, month, or day are outside of the normal range. This might be doable using the date-fns parse and isValid functions.

frontend/tests/utils/date-utils.test.ts
```diff
@@ -28,4 +29,36 @@ describe('date-utils', () => {
...
+      expect(isPastInTimezone(validTimeZone, date)).toEqual(true);
+    });
+  });
+
+  describe('isTodayInTimezone', () => {
+    const validTimeZone = 'Canada/Eastern';
+
+    const validDate = new Date();
+
+    const invalidDates = [addDays(new Date(), 1), addDays(new Date(), -1), '2025-02-20'];
```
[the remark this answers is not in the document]
gregory-j-baker: This doesn't address your problem.. but you should try to avoid using `new Date()` in unit tests because it leads to unpredictable tests.

Since the current instant in time is always changing, each time you run the test you will be using different data.

frontend/app/utils/date-utils.ts
faiza-jahanzeb: updated

frontend/tests/utils/date-utils.test.ts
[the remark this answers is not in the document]
faiza-jahanzeb: removed new Date() from tests

frontend/app/utils/date-utils.ts
gregory-j-baker: @sebastien-comeau --- I'm resolving this.

frontend/app/utils/date-utils.ts
```diff
@@ -69,3 +72,192 @@ export function isValidTimeZone(timeZone: string): boolean {
...
+ * Returns the start of the day in the specified timezone.
+ *
+ * This function takes a date (or timestamp) and a timezone, and returns a new Date object
+ * representing the start of the day (midnight) in that timezone.  If no date is provided,
+ * the current date and time are used.
+ *
+ * @param {string} timezone The IANA timezone string (e.g., 'America/New_York', 'Europe/London').
+ * @param {number | string | Date} [date] Optional date or timestamp to use. If not provided, the current date and time are used.
+ *                                      Can be a number (milliseconds since epoch), a string (parsable by `new Date()`), or a Date object.
+ * @returns {Date} A new Date object representing the start of the day in the specified timezone.
```
gregory-j-baker, line 201:
```suggestion
 * @param timezone - The IANA timezone string (e.g., 'America/New_York', 'Europe/London').
 * @param date - Optional date or timestamp to use. If not provided, the current date and time are used.
 *                                      Can be a number (milliseconds since epoch), a string (parsable by `new Date()`), or a Date object.
 * @returns A new Date object representing the start of the day in the specified timezone.
```

You don't have to declare the types in JSDoc when using typescript.

gregory-j-baker (approved):

PR #231, opened 2025-02-19
add protected/person-case/previous-sin route

Fbasham:
## Summary
[AB#16602](https://dev.azure.com/ESDCCM/a4b4f7af-bbd7-4f0d-83ab-a123d0b70fda/_workitems/edit/16602)

This PR adds:
- `/protected/person-case/previous-sin`
- `sin-utils.ts` for validating SINs using Luhn's algorithm
- unit tests for `sin-utils`

## Types of changes

## Checklist

- [x] linked this PR to a related issue (if applicable)

### DIFF
--- a/frontend/app/routes/protected/person-case/previous-sin.tsx
+++ b/frontend/app/routes/protected/person-case/previous-sin.tsx
@@ -1,0 +1,172 @@
+import { useId, useState } from 'react';
+import type { ChangeEvent } from 'react';
+
+import { data, useFetcher } from 'react-router';
+import type { RouteHandle, SessionData } from 'react-router';
+
+import { useTranslation } from 'react-i18next';
+import * as v from 'valibot';
+
+import type { Info, Route } from './+types/previous-sin';
+
+import { requireAuth } from '~/.server/utils/auth-utils';
+import { i18nRedirect } from '~/.server/utils/route-utils';
+import { Button } from '~/components/button';
+import { FetcherErrorSummary } from '~/components/error-summary';
+import { InputPatternField } from '~/components/input-pattern-field';
+import { InputRadios } from '~/components/input-radios';
+import { PageTitle } from '~/components/page-title';
+import { AppError } from '~/errors/app-error';
+import { getFixedT } from '~/i18n-config.server';
+import { handle as parentHandle } from '~/routes/protected/layout';
+import { getLanguage } from '~/utils/i18n-utils';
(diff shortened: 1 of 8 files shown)


frontend/app/.server/utils/sin-utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  return digitSum % 10 === 0;
+}
+
+/**
+ *
+ * @param sin - the Social Insurance Number (SIN)
+ * @returns a formatted SIN using a supplied separator
+ */
+export function formatSin(sin: string, separator = ' '): string {
+  if (!isValidSin(sin)) throw new Error('Invalid SIN format');
```
gregory-j-baker, line 64: Consider throwing an `AppError` here, instead. This will allow you to add an error code that will be displayed to the user, which can then be added to a remedy ticket for better troubleshooting. Ex:

``` ts
throw new AppError('Invalid SIN format', Errors.INVALID_SIN_FORMAT)
```

frontend/app/.server/utils/sin-utils.ts
gregory-j-baker, line 1: Might be a nitpick, but I don't think this file needs to be server-only, does it? I think it makes sense to move this out of `.server/` just in case we ever want to use it in the browser.

It should be tree-shaken if we aren't using it clientside, so there's no harm in moving it.

frontend/app/routes/protected/person-case/previous-sin.tsx
```diff
@@ -0,0 +1,173 @@
...
+  i18nNamespace: [...parentHandle.i18nNamespace, 'protected'],
+} as const satisfies RouteHandle;
+
+export async function loader({ context, request }: Route.LoaderArgs) {
+  requireAuth(context.session, new URL(request.url), ['user']);
+  const t = await getFixedT(request, handle.i18nNamespace);
+  return {
+    documentTitle: t('protected:previous-sin.page-title'),
+    defaultFormValues: context.session.inPersonSINCase?.previousSin,
+    sinInputPatternFormat,
```
gregory-j-baker, line 39: I don't think this needs to be returned from the loader. It shouldn't require any serverside code, so just add it to a file that's available on the client (see <https://github.com/DTS-STN/future-sir/pull/231/files#r1961888000>).

frontend/app/.server/utils/sin-utils.ts
Fbasham: I didn't even realize there was a `.server/utils` and `/utils`.  Fixed

PR #241, opened 2025-02-20
add countries.json and service to fetch localized countries

Fbasham:
## Summary
This PR aims to add PP country data which will be consumed by the frontend to display in a select input.  Currently this data is copy/paste from CDCP.  This PR also adds a 'service' layer for fetching localized country data (with the intention of having additional functions for other PP data).  I tried to implement the simplest approach, but I'm unsure how this would scale.  One potential drawback is that the data is fetched in the loader and the action (and the action really only needs the country IDs for validation purposes).  Let me know if it's a better approach to fetch all of the countries at the top level of form file, then get a localized list in the loader, or another approach that you think is more suitable for this problem. 

image

## Types of changes

### DIFF
--- a/frontend/app/.server/services/locale-data-service.ts
+++ b/frontend/app/.server/services/locale-data-service.ts
@@ -1,0 +1,27 @@
+import countryData from '~/.server/resources/countries.json';
+
+type Country = Readonly<{
+  id: string;
+  nameEn: string;
+  nameFr: string;
+}>;
+
+export function getCountries(): readonly Country[] {
+  return countryData.value.map((country) => ({
+    id: country.esdc_countryid,
+    nameEn: country.esdc_nameenglish,
+    nameFr: country.esdc_namefrench,
+  }));
+}
+
+type LocalizedCountry = Readonly<{
+  id: string;
+  name: string;
+}>;
+
+export function getLocalizedCountries(locale: Language = 'en'): readonly LocalizedCountry[] {
+  return getCountries().map((country) => ({
+    id: country.id,
+    name: country[locale === 'en' ? 'nameEn' : 'nameFr'],
+  }));
+}
--- a/frontend/app/routes/protected/person-case/contact-information.tsx
+++ b/frontend/app/routes/protected/person-case/contact-information.tsx
@@ -8,6 +8,7 @@ import * as v from 'valibot';
 
 import type { Info, Route } from './+types/contact-information';
 
+import { getLocalizedCountries } from '~/.server/services/locale-data-service';
 import { requireAuth } from '~/.server/utils/auth-utils';
 import { i18nRedirect } from '~/.server/utils/route-utils';
 import { Button } from '~/components/button';
@@ -26,18 +27,19 @@ type ContactInformationSessionData = NonNullable<SessionData['inPersonSINCase'][
 
 const VALID_LANGUAGES = ['english', 'french'] as const;
 
-const VALID_COUNTRIES = ['canada', 'other'] as const;
-
 export const handle = {
   i18nNamespace: [...parentHandle.i18nNamespace, 'protected'],
 } as const satisfies RouteHandle;
 
 export async function loader({ context, request }: Route.LoaderArgs) {
   requireAuth(context.session, new URL(request.url), ['user']);
+  const lang = getLanguage(request);
   const t = await getFixedT(request, handle.i18nNamespace);
+
   return {
     documentTitle: t('protected:contact-information.page-title'),
     defaultFormValues: context.session.inPersonSINCase?.contactInformation,
+    localizedCountries: getLocalizedCountries(lang),
   };
 }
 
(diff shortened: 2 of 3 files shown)


frontend/app/.server/services/locale-data-service.ts
```diff
@@ -0,0 +1,8 @@
+import countryData from '~/.server/resources/countries.json';
+
+export function getLocalizedCountries(locale = 'en') {
```
gregory-j-baker, line 3: In future-sir we've been trying to explicitly type the functions (where we can). Ex:

``` ts
type Country = {
  id: string;
  name: string;
}

function xxx(language: Language = 'en'): Country { .. }
```

Also, does it make sense to have a default value here? When would this ever be called without a language/locale parameter?

frontend/app/routes/protected/person-case/contact-information.tsx
```diff
@@ -26,18 +27,20 @@ type ContactInformationSessionData = NonNullable<SessionData['inPersonSINCase'][
...
 export async function loader({ context, request }: Route.LoaderArgs) {
   requireAuth(context.session, new URL(request.url), ['user']);
+  const lang = getLanguage(request);
   const t = await getFixedT(request, handle.i18nNamespace);
+  const localizedCountries = getLocalizedCountries(lang);
+
   return {
     documentTitle: t('protected:contact-information.page-title'),
     defaultFormValues: context.session.inPersonSINCase?.contactInformation,
+    localizedCountries,
```
gregory-j-baker, line 43: Why not just do

``` ts
localizedCountries: getLocalizedCountries(lang),
```

to save a line of code? I don't think storing this in a variable increases readability.

frontend/app/.server/services/locale-data-service.ts
Fbasham: Fair I guess, but isn't the return type automatically inferred by the code editor?  

The default value comes from the fact that `getLanguage()` can return `undefined`, so I explicitly set the default to `en` in the (impossible?) event that would ever happen.  

Anyways, I'll address these tomorrow.

gregory-j-baker: > isn't the return type automatically inferred by the code editor?

Yes, but our convention is to be explicit (in most cases). It requires the developer to be explicit and intentional, avoids accidental `any` types, and increases performance in large codebases.

Also, I 100% totally didn't ask ChatGPT for those points. 🏃🏻

frontend/app/routes/protected/person-case/contact-information.tsx
```diff
@@ -74,7 +76,10 @@ export async function action({ context, request }: Route.ActionArgs) {
         emailAddress: v.optional(
           v.pipe(v.string(), v.trim(), v.email(t('protected:contact-information.error-messages.email-address-invalid-format'))),
         ),
-        country: v.picklist(VALID_COUNTRIES, t('protected:contact-information.error-messages.country-required')),
+        country: v.picklist(
+          getLocalizedCountries(lang).map(({ id }) => id),
```
sebastien-comeau, line 80: You don't have to get localized contries in this context.

```suggestion
          getCountries().map(({ id }) => id),
```

sebastien-comeau (approved):

frontend/app/.server/services/locale-data-service.ts
gregory-j-baker: > The default value comes from the fact that getLanguage() can return undefined

I think you're looking at this from the top-down, when I think it makes more sense to  approach it from the bottom-up.

Calees (in this case `getLocalizedCountries()`) shouldn't adapt to the caller; instead, the callers should conform to the callee's expectations.

That said, if you disagree, you can disregard my suggestion. It's not a critical issue, and I don't have strong feelings about it in this case.

sebastien-comeau: I started with the belief that everything should be explicitly typed. Then, I shifted to thinking implicit types were better, allowing inference to handle it. But then I realized—being explicit is actually preferable because it ensures the function logic aligns with the return type and helps developers understand the function's intent just by reading its signature.

Fbasham: I guess it stems from the annoyance of having to do `getLocalizedCountries(lang ?? 'en')` everywhere you'd want to call the function.  Not really a big deal, but 2 schools of though I guess.

gregory-j-baker: ChatGPT gave many reasons why you should be explicit with your return types, all of which make sense to me.

> When refactoring, if you rely on inferred types, TypeScript may silently change the inferred return type, leading to subtle bugs. Explicit return types ensure the contract remains the same.

Also, sometimes we loosen or tighten the types depending on intent.

gregory-j-baker: > I guess it stems from the annoyance of having to do `getLocalizedCountries(lang ?? 'en')` everywhere you'd want to call the function. Not really a big deal, but 2 schools of though I guess.

I have a PR awaiting approval that will remove `undefined` when fetching a language.

see [#245](https://github.com/DTS-STN/future-sir/pull/245/files#diff-a6630785c822cc63bda3136c76f6f77d06e7d7d594f7983411c177707e857f86R53)

Fbasham: ok cool cool.  I'm going to merge this then and refactor since I have more PP data to load and subsequently more changes to make.

sebastien-comeau: See the function implementation signature as if you were first defining the type. Creating the function type enforces a structured function contract. Remember, our job involves more reading than writing code. While an IDE can infer types easily, without it, they aren't immediately clear.

frontend/app/routes/protected/person-case/contact-information.tsx
Fbasham: will make this change in a separate PR since there's more PP data to load for this route.

frontend/app/.server/services/locale-data-service.ts
gregory-j-baker: > See the function implementation signature as if you were first defining the type. Creating the function type enforces a structured function contract. Remember, our job involves more reading than writing code. While an IDE can infer types easily, without it, they aren't immediately clear.

The more you know 🌈

PR #274, opened 2025-02-25
feat(frontend): Parent Details Screen

Dario-Au:
## Summary

Added Parent Details Screen
[AB#16028](https://dev.azure.com/ESDCCM/a4b4f7af-bbd7-4f0d-83ab-a123d0b70fda/_workitems/edit/16028)

## Types of changes

## Checklist

<details>

  ``` shell
  npm run lint:check
  npm run format:check
  ```
</details>

<details>

  ``` shell
  npm run test
  npm run test:e2e
  ```
</details>

### DIFF
--- a/frontend/app/routes/protected/person-case/parent-details.tsx
+++ b/frontend/app/routes/protected/person-case/parent-details.tsx
@@ -20,7 +20,6 @@ import { InputSelect } from '~/components/input-select';
 import { PageTitle } from '~/components/page-title';
 import { AppError } from '~/errors/app-error';
 import { ErrorCodes } from '~/errors/error-codes';
-import { useLanguage } from '~/hooks/use-language';
 import { getTranslation } from '~/i18n-config.server';
 import { handle as parentHandle } from '~/routes/protected/layout';
 import { REGEX_PATTERNS } from '~/utils/regex-utils';
@@ -31,24 +30,33 @@ type ParentDetailsSessionData = NonNullable<SessionData['inPersonSINCase']['pare
 const COUNTRIES = ['CAN', 'FRA'] as const;
 const PROVINCES = ['AB', 'BC', 'MB', 'NB', 'NL', 'NT', 'NS', 'NU', 'ON', 'PE', 'QC', 'SK', 'YT'] as const;
 
+const COUNTRY_CODE_CANADA = 'CAN';
+const MAX_PARENTS = 4;
+
 export const handle = {
   i18nNamespace: [...parentHandle.i18nNamespace, 'protected'],
 } as const satisfies RouteHandle;
 
 export async function loader({ context, request }: Route.LoaderArgs) {
   requireAuth(context.session, new URL(request.url), ['user']);
+
   const { t } = await getTranslation(request, handle.i18nNamespace);
+  const sessionData = context.session.inPersonSINCase?.parentDetails ?? [];
+
   return {
     documentTitle: t('protected:parent-details.page-title'),
-    defaultFormValues:
-      context.session.inPersonSINCase?.parentDetails?.map((details) => ({
-        unavailable: details.unavailable,
-        givenName: !details.unavailable ? details.givenName : undefined,
-        lastName: !details.unavailable ? details.lastName : undefined,
-        country: !details.unavailable ? details.birthLocation.country : undefined,
-        province: !details.unavailable ? details.birthLocation.province : undefined,
-        city: !details.unavailable ? details.birthLocation.city : undefined,
-      })) ?? [],
+    defaultFormValues: sessionData.map((details) =>
+      details.unavailable
+        ? { unavailable: true }
+        : {
+            unavailable: false,
+            givenName: details.givenName,
+            lastName: details.lastName,
+            country: details.birthLocation.country,
+            province: details.birthLocation.province,
+            city: details.birthLocation.city,
+          },
+    ),
   };
 }
 
@@ -63,8 +71,6 @@ export async function action({ context, request }: Route.ActionArgs) {
   const formData = await request.formData();
   const action = formData.get('action');
   const maxStringLength = 100;
-  const maxParents = 4;
-  const canadaCountryCode = 'CAN';
 
   switch (action) {
     case 'back': {
@@ -100,7 +106,7 @@ export async function action({ context, request }: Route.ActionArgs) {
                   'country',
                   [
                     v.object({
-                      country: v.literal(canadaCountryCode, t('protected:parent-details.country-error.invalid-country')),
+                      country: v.literal(COUNTRY_CODE_CANADA, t('protected:parent-details.country-error.invalid-country')),
                       province: v.pipe(
                         v.string(t('protected:parent-details.province-error.required-province')),
                         v.trim(),
@@ -120,7 +126,7 @@ export async function action({ context, request }: Route.ActionArgs) {
                       country: v.pipe(
                         v.string(t('protected:parent-details.country-error.required-country')),
                         v.nonEmpty(t('protected:parent-details.country-error.required-country')),
-                        v.excludes(canadaCountryCode, t('protected:parent-details.country-error.invalid-country')),
+                        v.excludes(COUNTRY_CODE_CANADA, t('protected:parent-details.country-error.invalid-country')),
                         v.picklist(COUNTRIES, t('protected:parent-details.country-error.invalid-country')),
                       ),
                       province: v.optional(
@@ -152,11 +158,11 @@ export async function action({ context, request }: Route.ActionArgs) {
           t('protected:parent-details.details-unavailable'),
         ),
         v.minLength(1),
-        v.maxLength(maxParents),
+        v.maxLength(MAX_PARENTS),
       ) satisfies v.GenericSchema<ParentDetailsSessionData>;
 
       const parentAmount = Number(formData.get('parent-amount')) || 0;
-      const inputLength = Math.min(parentAmount, maxParents);
+      const inputLength = Math.min(parentAmount, MAX_PARENTS);
 
       const input = Array.from({ length: inputLength }).map((_, i) => ({
         unavailable: Boolean(formData.get(`${i}-unavailable`)),
@@ -187,8 +193,10 @@ export async function action({ context, request }: Route.ActionArgs) {
 
 export default function CreateRequest({ loaderData, actionData, params }: Route.ComponentProps) {
   const { t } = useTranslation(handle.i18nNamespace);
+
   const fetcherKey = useId();
   const fetcher = useFetcher<Info['actionData']>({ key: fetcherKey });
+
   const isSubmitting = fetcher.state !== 'idle';
   const errors = fetcher.data?.errors;
   const defaultFormValues = loaderData.defaultFormValues;
@@ -231,16 +239,17 @@ function ParentInformation({ defaultFormValues, errors }: ParentInformationProps
   const { t } = useTranslation(handle.i18nNamespace);
   const { idList, addId, removeId } = useIdList(Math.max(defaultFormValues.length, 1));
 
-  const maxParents = 4;
-  const canAddParent = idList.length < maxParents;
+  const canAddParent = idList.length < MAX_PARENTS;
 
-  const onAddParent = () => {
+  function onAddParent() {
     if (canAddParent) addId();
-  };
-  const onRemoveParent = (index: number) => {
-    defaultFormValues.splice(index, 1); //Remove autofill for the removed parent
+  }
+
+  function onRemoveParent(index: number) {
+    // remove parent data from the form values
+    defaultFormValues.splice(index, 1);
     removeId(index);
-  };
+  }
 
   return (
     <>
@@ -252,7 +261,7 @@ function ParentInformation({ defaultFormValues, errors }: ParentInformationProps
             index={index}
             defaultValues={defaultFormValues[index]}
             errors={errors}
-            onClose={idList.length > 1 ? onRemoveParent : undefined}
+            onRemove={idList.length > 1 ? onRemoveParent : undefined}
           />
         ))}
       </div>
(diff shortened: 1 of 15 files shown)


frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+      const parseResult = v.safeParse(schema, input, { lang });
+
+      if (!parseResult.success) {
+        return data({ errors: v.flatten<typeof schema>(parseResult.issues).nested }, { status: 400 });
+      }
+
+      context.session.inPersonSINCase ??= {};
+      context.session.inPersonSINCase.parentDetails = parseResult.output;
+
+      throw i18nRedirect('routes/protected/person-case/previous-sin.tsx', request);
```
gregory-j-baker, line 180:
```suggestion
      (context.session.inPersonSINCase ??= {}).parentDetails = parseResult.output;
      throw i18nRedirect('routes/protected/person-case/previous-sin.tsx', request);
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    default: {
+      throw new AppError(`Unrecognized action: ${action}`);
```
gregory-j-baker, line 184:
```suggestion
      throw new AppError(`Unrecognized action: ${action}`, ErrorCodes.UNRECOGNIZED_ACTION);
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+export async function action({ context, request }: Route.ActionArgs) {
+  requireAuth(context.session, new URL(request.url), ['user']);
+
+  const { lang, t } = await getTranslation(request, handle.i18nNamespace);
+  const formData = await request.formData();
+  const action = formData.get('action');
+  const maxStringLength = 100;
+  const maxParents = 4;
+  const canadaCountryCode = 'CAN';
```
gregory-j-baker, line 64: Do these need to be variables? I think inlining these where they're used would be acceptable.

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+            t('protected:parent-details.details-unavailable'),
+          ),
+          t('protected:parent-details.details-unavailable'),
+        ),
+        v.minLength(1),
+        v.maxLength(maxParents),
+      ) satisfies v.GenericSchema<ParentDetailsSessionData>;
+
+      const parentAmount = Number(formData.get('parent-amount')) || 0;
+      const inputLength = parentAmount > maxParents ? maxParents : parentAmount;
```
gregory-j-baker, line 156:
```suggestion
      const inputLength = Math.min(parentAmount, maxParents);
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+          unavailable: Boolean(formData.get(`${i}-unavailable`)),
+          givenName: String(formData.get(`${i}-given-name`)),
+          lastName: String(formData.get(`${i}-last-name`)),
+          birthLocation: {
+            country: String(formData.get(`${i}-country`)),
+            province: String(formData.get(`${i}-province`)),
+            city: String(formData.get(`${i}-city`)),
+          },
+        };
+      }) satisfies Partial<ParentDetailsSessionData>;
```
gregory-j-baker, line 169:
```suggestion
      const input = Array.from({ length: inputLength }).map((_, i) => ({
        unavailable: Boolean(formData.get(`${i}-unavailable`)),
        givenName: String(formData.get(`${i}-given-name`)),
        lastName: String(formData.get(`${i}-last-name`)),
        birthLocation: {
          country: String(formData.get(`${i}-country`)),
          province: String(formData.get(`${i}-province`)),
          city: String(formData.get(`${i}-city`)),
        },
      })) satisfies ParentDetailsSessionData;
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+}
+
+export default function CreateRequest({ loaderData, actionData, params }: Route.ComponentProps) {
+  const { t } = useTranslation(handle.i18nNamespace);
+  const fetcherKey = useId();
+  const fetcher = useFetcher<Info['actionData']>({ key: fetcherKey });
+  const isSubmitting = fetcher.state !== 'idle';
+  const errors = fetcher.data?.errors;
+  const defaultFormValues = loaderData.defaultFormValues;
+  const [parentAmount, setParentAmount] = useState<number>(defaultFormValues?.length ?? 1);
```
gregory-j-baker, line 196: You don't have to explicitly type the `useState()` call when it can be inferred from the initial value.


```suggestion
  const [parentAmount, setParentAmount] = useState(defaultFormValues?.length ?? 1);
```

frontend/app/routes/protected/person-case/parent-details.tsx
Dario-Au: I feel it makes it easier to read and update in the future if necessary

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+}
+
+type detailsType = {
+  unavailable?: boolean;
+  givenName?: string;
+  lastName?: string;
+  country?: string;
+  province?: string;
+  city?: string;
+};
```
gregory-j-baker, line 231:
```suggestion
type FormData = {
  unavailable?: boolean;
  givenName?: string;
  lastName?: string;
  country?: string;
  province?: string;
  city?: string;
};
```

Types are typically defined with their first character uppercased. And I don't think `details` is a good description for this type. As far as I can tell, it represents the values in the form, so why not call it `FormData` or something that fits more with what the data represents?

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+        ))}
+      </div>
+      {canAddParent && (
+        <Button size="lg" type="button" variant="link" endIcon={faPlus} className="mt-3 px-3" onClick={onAddParent}>
+          {t('protected:parent-details.add-parent')}
+        </Button>
+      )}
+    </>
+  );
+}
```
gregory-j-baker, line 275: I'm really struggling to read and understand this code.. When looking at this component, it's not clear to me what `amount` means. Amount of what?
Why is `parentAmount` set to zero or one depending on the value of `amount`? What does `setAmount(parentAmount + 1)` do?
How does the `onClose` handler work?

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+interface ParentFormProps {
+  index: number;
+  defaultValues?: detailsType;
+  errors?: Record<string, [string, ...string[]] | undefined>;
+  onClose?: () => void;
+}
+
+function ParentForm({ index, defaultValues, errors, onClose }: ParentFormProps) {
+  const { t } = useTranslation(handle.i18nNamespace);
+  const [unavailable, setUnavailable] = useState<boolean>(defaultValues?.unavailable ?? false);
```
gregory-j-baker, line 286:
```suggestion
  const [unavailable, setUnavailable] = useState(defaultValues?.unavailable ?? false);
```

Also, I think this variable should be named `parentDetailsUnavailable`. I had to run the application and navigate to the page to understand what `unavailable` was being used for.

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+  const [country, setCountry] = useState<string | undefined>(defaultValues?.country);
```
gregory-j-baker, line 287:
```suggestion
  const [country, setCountry] = useState(defaultValues?.country);
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+  const countryOptions = (['select-option', ...COUNTRIES] as const).map((value) => ({
+    value: value === 'select-option' ? '' : value,
+    children: t(`protected:countries.${value}`),
+  }));
+
+  const provinceOptions = (['select-option', ...PROVINCES] as const).map((value) => ({
+    value: value === 'select-option' ? '' : value,
+    children: t(`protected:provinces.${value}`),
+  }));
```
gregory-j-baker, line 298:
```suggestion
  const provinceOptions = (['select-option', ...PROVINCES] as const)
    .map((value) => ({
      value: value === 'select-option' ? '' : value,
      children: t(`protected:provinces.${value}`),
    }))
    .sort(({ children: left }, { children: right }) => left.localeCompare(right, lang));
```

gregory-j-baker: Sorry.. I added some comments in vscode using the github extension and it seems to have messed up a bunch of the line numbers.

Anyway, the idea here is to sort the `<select>` by province name, using the correct locale.

gregory-j-baker: I just realized that the code I suggested doesn't move the "select option" option to the top.

gregory-j-baker:
``` ts
    .sort((left, right) => {
      if (left.value === '') return -1; // sort empty string to the top
      if (right.value === '') return 1; // sort empty string to the top
      return left.children.localeCompare(right.children, lang);
    });
```

frontend/app/routes/protected/person-case/parent-details.tsx
Dario-Au: - `amount` is the base amount of parents to render, this number can be 0
- `parentAmount` is just `amount` with a minimum of 1
  - The component needs to render at least one parent form by default 
  - I'll replace `amount > 0 ? amount : 1;` with  `Math.max(amount, 1)`
- `setAmount(parentAmount + 1)` increases `amount` which in turn decreases the amount of parents to render
- `onCloseParent`  decreases `amount` which in turn decreases the amount of parents to render

gregory-j-baker: Your explanation makes perfect sense, and I understand the component now that I have the context. However, without that context, I just couldn't figure it out.. I gave up trying after about 10 minutes.

Could the code be refactored to be more self-documenting in support of this explanation?

For example, maybe `amount` could be renamed to `currentNumberOfParents` or `nCurrentParents` or something like that.

Maybe rename the `onAddParent` to just `addParent` and `onCloseParent` to `removeParent`. 

If the code can't be refactored to be self-documenting, then at least add a JSDoc comment explaining what the component does. I really struggled with this one and after 10 minutes of trying to figure it out, just gave up.

gregory-j-baker:
Is it intentional that you can only remove the last-added parent? What if you want to remove the first one after you've added four? That seems like it could be a bit of a nuisance for the user.

frontend/app/routes/protected/person-case/parent-details.tsx
gregory-j-baker: I just noticed the use of `?? false` here. This is effectively setting the default state to `false`, so why not just return `false` for this property from the loader?

Is there a reason you return `undefined` and then do a `?? false` during render, or was that just an oversight?

Sorry.. I find react code incredibly difficult to reason about sometimes.

frontend/app/routes/protected/person-case/parent-details.tsx
gregory-j-baker: I was thinking a bit more about this and I think instead of passing in `setAmount()`, you should do something like this:

```
function (..., addParentFn, removeParentFn) {
  // handlers
  function addParent() {
    // code required by this component (if any)
    if (addParentFn) addParentFn();
  }

  function removeParent() {
    // core required by this component (if any)
    if (removeParentFn) removeParentFn();
  }
}
```

Allow the calling/wrapping component to define what happens when you add or remove a parent.

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,424 @@
...
+    <>
+      <input type="hidden" name="parent-amount" value={parentAmount} />
+      <div className="space-y-10">
+        {Array.from({ length: parentAmount }).map((_, index) => (
+          <ParentForm
+            key={getId(index)}
+            index={index}
+            defaultValues={defaultFormValues[index]}
+            errors={errors}
+            onClose={parentAmount > 1 ? onCloseParent : undefined}
```
gregory-j-baker, line 257: I don't like the name `onCloseParent` here. It really doesn't describe what is happening (in my opinion). You're not closing a parent, you're removing it, so it should be `removeParent`. Also, the index should be passed in here so it is clear what is happening.

I think it would be better to use something like:

``` tsx
onClose={parentAmount > 1 ? () => removeParent(index) : undefined}
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,424 @@
...
+        </>
+      )}
+    </div>
+  );
+}
+
+/*
+  Manages unique Ids based on the provided indexes from a list.
+*/
+const useIdList = () => {
```
gregory-j-baker, line 393:
```suggestion
function useIdList() { .. }
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,424 @@
...
+  const { getId, removeId } = useIdList();
+
+  const maxParents = 4;
+  const canAddParent = parentAmount < maxParents;
+
+  const onAddParent = () => {
+    if (canAddParent) setParentAmount(parentAmount + 1);
+  };
+  const onCloseParent = (index: number) => {
+    defaultFormValues.splice(index, 1);
```
gregory-j-baker, line 242: What is this line doing? I *think* it's unnecessary.

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,424 @@
...
+    removeId(index);
+    setParentAmount(parentAmount - 1);
```
gregory-j-baker, line 244:
```suggestion
    setParentAmount((prev) => prev - 1);
```

frontend/app/routes/protected/person-case/parent-details.tsx
```diff
@@ -0,0 +1,424 @@
...
+     * @param index - The index from a list.
+     * @returns - A unique Id corresponding to the index.
+     */
+    getId: (index: number): number => {
+      if (idList[index]) return idList[index];
+      const newId = id + 1;
+      setId(newId);
+      setIdList((prev) => [...prev, newId]);
+      return newId;
+    },
```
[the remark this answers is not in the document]
gregory-j-baker: Also, using two separate states for this is unnecessary. Why not just use a `Map<number, number>` and some random string generator? 

Do the IDs have to be sequential? If so, the code might be fragile.

frontend/app/routes/protected/person-case/parent-details.tsx
Dario-Au: This prevents autofill on removed parents
1. Lets say `defaultFormValues` autofills the 1st and 2nd parent
2. I then add a new parent so there are current 3 parents
   1. Autofilled parent
   2. Autofilled parent
   3. New parent
3. I then remove the 2nd parent, if `defaultFormValues` isn't updated, it will autofill the previously 3rd parent
   1. Autofilled parent
   2. New parent <- `defaultFormValues[1]` attempts to autofill

gregory-j-baker: My bad, I didn't realize that `.splice()` was a *mutative* operation. Man, JS is really bad sometimes.

I think you should add a comment to this line explaining what it's doing.

PR #295, opened 2025-02-27
feat(frontend): display secondary docs and current name info on review screen

faiza-jahanzeb:
## Summary

[AB#14685](https://dev.azure.com/ESDCCM/a4b4f7af-bbd7-4f0d-83ab-a123d0b70fda/_workitems/edit/14685)
---- Add secondary docs and current name information on review screen

## Types of changes

## Checklist

<details>

  ``` shell
  npm run lint:check
  npm run format:check
  ```
</details>

<details>

  ``` shell
  npm run test
  npm run test:e2e
  ```
</details>

## Additional Notes

If this PR introduces significant changes, explain your reasoning and provide any necessary context here. Feel free to include diagrams, screenshots, or alternative approaches you considered.

## Screenshots (if applicable)
For the Preferred name, display the current first name and current last name from primary doc information
image
For the Preferred name, display the current name , current middle name and current last name from the current name information
preferred name

### DIFF
--- a/frontend/app/routes/protected/person-case/review.tsx
+++ b/frontend/app/routes/protected/person-case/review.tsx
@@ -8,11 +8,12 @@ import { useTranslation } from 'react-i18next';
 
 import type { Info, Route } from './+types/review';
 
+import { applicantGenderService } from '~/.server/domain/person-case/services';
 import { requireAuth } from '~/.server/utils/auth-utils';
 import { i18nRedirect } from '~/.server/utils/route-utils';
 import { Button } from '~/components/button';
-import { ButtonLink } from '~/components/button-link';
 import { DescriptionList, DescriptionListItem } from '~/components/description-list';
+import { InlineLink } from '~/components/links';
 import { PageTitle } from '~/components/page-title';
 import { AppError } from '~/errors/app-error';
 import { ErrorCodes } from '~/errors/error-codes';
@@ -29,10 +30,20 @@ export async function loader({ context, request }: Route.LoaderArgs) {
   const tabId = new URL(request.url).searchParams.get('tid');
   if (!tabId) throw new AppError('Missing tab id', ErrorCodes.MISSING_TAB_ID, { httpStatusCode: 400 });
 
-  const { t } = await getTranslation(request, handle.i18nNamespace);
+  const { t, lang } = await getTranslation(request, handle.i18nNamespace);
   const inPersonSINCase = validateInPersonSINCaseSession(context.session, tabId, request);
 
-  return { documentTitle: t('protected:review.page-title'), inPersonSINCase, tabId };
+  return {
+    documentTitle: t('protected:review.page-title'),
+    inPersonSINCase: {
+      ...inPersonSINCase,
+      primaryDocuments: {
+        ...inPersonSINCase.primaryDocuments,
+        genderName: applicantGenderService.getLocalizedApplicantGenderById(inPersonSINCase.primaryDocuments.gender, lang).name,
+      },
+    },
+    tabId,
+  };
 }
 
 function validateInPersonSINCaseSession(
@@ -176,13 +187,12 @@ export default function Review({ loaderData, actionData, params }: Route.Compone
       <div className="max-w-prose">
         <fetcher.Form method="post" noValidate>
           <p className="mb-8 text-lg">{t('protected:review.read-carefully')}</p>
-          <div className="space-y-10 divide-y border-y">
+          <div className="space-y-10">
             <section className="space-y-6">
-              <h2 className="font-lato mt-3 text-2xl font-bold">{t('protected:primary-identity-document.page-title')}</h2>
-              <DescriptionList>
+              <h2 className="font-lato text-2xl font-bold">{t('protected:primary-identity-document.page-title')}</h2>
+              <DescriptionList className="divide-y border-y">
                 <DescriptionListItem term={t('protected:primary-identity-document.current-status-in-canada.title')}>
                   <p>
-                    {/* TODO: Code Table Value  */}
                     {t(
                       `protected:primary-identity-document.current-status-in-canada.options.${inPersonSINCase.primaryDocuments.currentStatusInCanada}` as 'protected:primary-identity-document.current-status-in-canada.options.select-option',
                     )}
@@ -190,7 +200,6 @@ export default function Review({ loaderData, actionData, params }: Route.Compone
                 </DescriptionListItem>
                 <DescriptionListItem term={t('protected:primary-identity-document.document-type.title')}>
                   <p>
-                    {/* TODO: Code Table Value  */}
                     {t(
                       `protected:primary-identity-document.document-type.options.${inPersonSINCase.primaryDocuments.documentType}` as 'protected:primary-identity-document.document-type.options.select-option',
                     )}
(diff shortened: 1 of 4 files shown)


frontend/app/routes/protected/person-case/review.tsx
```diff
@@ -200,6 +198,90 @@ export default function Review({ loaderData, actionData, params }: Route.Compone
...
+                      ? inPersonSINCase.primaryDocuments.givenName
+                      : inPersonSINCase.currentNameInfo.firstName}
+                  </p>
+                </DescriptionListItem>
+                <DescriptionListItem term={t('protected:current-name.preferred-name.middle-name')}>
+                  <p>
+                    {inPersonSINCase.currentNameInfo.preferredSameAsDocumentName
+                      ? ''
+                      : inPersonSINCase.currentNameInfo.middleName}
+                  </p>
```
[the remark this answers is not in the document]
faiza-jahanzeb: You mean if the applicant's preferred name same as their primary identity document name? is Yes, then we don't need to display the rest of the fields? 
image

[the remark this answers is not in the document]
faiza-jahanzeb: Or we would still display the values for Current first name and Current last name from primary document?

sebastien-comeau (changes requested):
Trivial changes to make:

- `<div className="space-y-10 divide-y border-y">` changed to `<div className="space-y-10">`
- `<DescriptionList>` changed to `<DescriptionList className="divide-y border-y">`
- Use `InlineLink` component instead of `ButtonLink`

frontend/app/routes/protected/person-case/review.tsx
```diff
@@ -196,9 +199,89 @@ export default function Review({ loaderData, actionData, params }: Route.Compone
...
+                <DescriptionListItem term={t('protected:review.document-uploaded')}>
+                  <p>{t('protected:review.choosen-file')}</p>
+                </DescriptionListItem>
+              </DescriptionList>
+              <InlineLink file="routes/protected/person-case/secondary-doc.tsx">
+                {t('protected:review.edit-secondary-identity-document')}
+              </InlineLink>
+            </section>
+            <section className="space-y-6">
+              <h2 className="font-lato mt-3 text-2xl font-bold">{t('protected:review.sub-title-preferred-name')}</h2>
```
sebastien-comeau, line 228: Headings #2 don't need `mt-3` class name since the section are spaced from their parent element with class `space-y-10`.

frontend/app/routes/protected/person-case/review.tsx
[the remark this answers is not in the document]
gregory-j-baker: The comment was more about not rendering `<p></p>` when a value is not present.

frontend/app/routes/protected/person-case/review.tsx
```diff
@@ -145,27 +149,30 @@ export default function Review({ loaderData, actionData, params }: Route.Compone
   const fetcher = useFetcher<Info['actionData']>({ key: fetcherKey });
   const isSubmitting = fetcher.state !== 'idle';
 
+  function getLocalizedGenderById(id: string): string | undefined | null {
+    const gender = loaderData.localizedGenders.find((g) => g.id == id);
+    return gender?.name;
+  }
```
gregory-j-baker, line 155: Are you able to use this function to get what you need (probably calling it from the loader)?

<https://github.com/DTS-STN/future-sir/blob/main/frontend/app/.server/services/locale-data-service.ts#L315>

faiza-jahanzeb: no, how do I get the language inside the function Review?  const { t } = useTranslation(handle.i18nNamespace); doesn't have lang

gregory-j-baker: In the loader you can call:

```ts
const { lang, t } = await getTranslation(request, handle.i18nNamespace);
```

Then just fetch and return the gender you're interested in back to the route component.

It's a minor optimization, for sure.. but I think it's good practice to only send the data that you need to render. Sending all the genders, only to pick a single one at render time is a form of overfetching.

Still, if you think this suggestion will complicate the code or impact code readability, feel free to just mark this comment as resolved and ignore it. It's not really a critical issue.

faiza-jahanzeb: next we have country and provinces data also on this page, so it's better to utilize what we already have in locale-data-service... For the rest of the elements inside primaryDocuments, how would we access them ? eg like this: `citizenshipDate: String(inPersonSINCase.primaryDocuments?.citizenshipDate) `or like this: `citizenshipDate: inPersonSINCase.primaryDocuments?.citizenshipDate ?? ''`

gregory-j-baker: Using `String()` isn't recommended because it will convert nulls and undefined values to strings:

```
➜  ~ node                                                                    
Welcome to Node.js v22.14.0.
Type ".help" for more information.
> String(null)
'null'
> String(undefined)
'undefined'
```

You should use conditional rendering to suppress rendering when these values don't exist.

frontend/app/routes/protected/person-case/review.tsx
```diff
@@ -25,9 +26,20 @@ export const handle = {
...
+    inPersonSINCase,
+  };
+}
+
+function toLocalizedGender(genderId: string | undefined, locale: Language = 'en'): string {
+  try {
+    return genderId ? (getLocalizedApplicantGenderById(genderId, locale).name ?? '') : '';
+  } catch {
+    return '';
+  }
```
gregory-j-baker, line 42: Is catching and then returning an empty string correct here? It seems to me like you should probably let the `getLocalizedApplicantGenderById()` call throw if the gender id can't be resolved.

Something like this:

```ts
function toLocalizedGender(genderId: string | undefined, locale: Language): string {
  if (!genderId) return '';
￼￼return getLocalizedApplicantGenderById(genderId, locale).name;
}
```

I feel like absorbing the error (by catching and returning an empty string) can lead to hidden bugs down the road.

frontend/app/routes/protected/person-case/review.tsx
```diff
@@ -39,6 +51,7 @@ export async function loader({ context, request }: Route.LoaderArgs) {
 function validateInPersonSINCaseSession(
   session: AppSession,
   request: Request,
+  locale: Language = 'en',
```
gregory-j-baker, line 54: I don't think this should have a default value. There are no times when this function should ever be called without a valid language.

also:

```suggestion
  language: Language,
```

Reason: `'en'` and `'fr'` are not locales.

frontend/app/routes/protected/person-case/review.tsx
faiza-jahanzeb: this would return string or null ```
function toLocalizedGender(genderId: string | undefined, language: Language): string | null {
  if (!genderId) return '';
  return getLocalizedApplicantGenderById(genderId, language).name;
}
```
and here we'll be null coalescing the gender:
```
primaryDocuments: {
   gender: toLocalizedGender(inPersonSINCase.primaryDocuments?.gender, locale) ?? '',
},
```that would be fine I guess?

sebastien-comeau: Localized applicant gender should be resolved and return by the loader using `getLocalizedX` method.

faiza-jahanzeb: updated

sebastien-comeau: I've refactor your latest changes locally to move `gender` name resolution inside `loader`. `validateInPersonSINCaseSession` should only responsible to validate the session state.

https://github.com/DTS-STN/future-sir/commit/d1b7586373ef59d81cf86f1bca80d00d7fef1035

sebastien-comeau (approved):

PR #300, opened 2025-02-28
feat(frontend): generate a browser tabid per flow

gregory-j-baker:
## Summary

This PR lays the groundwork for a state machine managing the in-person application flows by introducing `tabId` tracking. This feature will eventually allow users to initiate and complete multiple flows concurrently in separate browser tabs without session data conflicts.

## Problem

Currently, session data is shared across all browser tabs, which prevents users from completing multiple in-person application flows concurrently.

## Solution

To allow users to complete multiple in-person application flows concurrently, we need a way to isolate the data that is stored in session for each browser tab. The best way to handle this is to leverage the browser's built-in session store feature to store a unique id, which is then added to all requests in the flow.

## Changes in this PR

- Ensure the `tabId` is propagated throughout the relevant parts of the in-person case flow.

# Future work

This PR is a prerequisite for future work implementing tab-specific session data management and the state machine. Subsequent PRs will:

* Utilize the `tabId` to isolate session data.
* Introduce a state machine to manage the flow of the in-person application.

## Types of changes

## Checklist

### DIFF
--- a/frontend/app/errors/error-codes.ts
+++ b/frontend/app/errors/error-codes.ts
@@ -13,6 +13,7 @@ export const ErrorCodes = {
   MISSING_LANG_PARAM: 'CMP-0001',
 
   // form error codes
+  MISSING_TAB_ID: 'FRM-0000',
   UNRECOGNIZED_ACTION: 'FRM-0001',
 
   // i18n error codes
--- a/frontend/app/i18n-routes.ts
+++ b/frontend/app/i18n-routes.ts
@@ -104,91 +104,105 @@ export const i18nRoutes = [
         paths: { en: '/en/protected/request', fr: '/fr/protege/requete' },
       },
       {
-        id: 'PROT-0004',
-        file: 'routes/protected/person-case/privacy-statement.tsx',
-        paths: {
-          en: '/en/protected/person-case/privacy-statement',
-          fr: '/fr/protege/cas-personnel/declaration-de-confidentialite',
-        },
-      },
-      {
-        id: 'PROT-0005',
-        file: 'routes/protected/person-case/primary-docs.tsx',
-        paths: { en: '/en/protected/person-case/primary-documents', fr: '/fr/protege/cas-personnel/documents-primaires' },
-      },
-      {
-        id: 'PROT-0006',
-        file: 'routes/protected/person-case/request-details.tsx',
-        paths: { en: '/en/protected/person-case/request-details', fr: '/fr/protege/cas-personnel/faire-une-demande' },
-      },
-      {
-        id: 'PROT-0007',
-        file: 'routes/protected/person-case/current-name.tsx',
-        paths: {
-          en: '/en/protected/person-case/current-name',
-          fr: '/fr/protege/cas-personnel/nom-actuel',
-        },
-      },
-      {
-        id: 'PROT-0008',
-        file: 'routes/protected/person-case/personal-info.tsx',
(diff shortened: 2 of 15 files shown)


sebastien-comeau (commented):
Wouldn't be simpler to generate and use a  [URL-friendly unique string ID](https://www.npmjs.com/package/nanoid) in URL param like CDCP? I don't say you solution is wrong but it seems very complicated for what it does.

**Example:** /en/protected/person-case/**0m4gksf6fBf**/overview

gregory-j-baker:
> Wouldn't be simpler to generate and use a [URL-friendly unique string ID](https://www.npmjs.com/package/nanoid) in URL param like CDCP? I don't say you solution is wrong but it seems very complicated for what it does.
> 
> **Example:** /en/protected/person-case/create/**0m4gksf6fBf**/overview

The CDCP implementation doesn't work. In the video below I have the same flow open in two different tabs, clobbering the data in one with the data in the other.

[Screencast from 2025-02-28 07-37-12.webm](https://github.com/user-attachments/assets/467a0a85-abaa-4cd9-b111-59db68c437fe)

sebastien-comeau:
> The CDCP implementation doesn't work. In the video below I have the same flow open in two different tabs, clobbering the data in one with the data in the other.

I understand that, but another concern is that since we don’t generate a temporary state ID within the same tab, a user could start creating a case, navigate elsewhere in the application midway, then return later to create a new case—only to find the previously entered data still present. We need to prevent multiple case creations from conflicting within the same tab. That’s why using a URL parameter with a unique identifier makes sense, as the tab ID remains constant within the same tab.

> rephrased by AI

gregory-j-baker:
> ...return later to create a new case—only to find the previously entered data still present...
> ...using a URL parameter with a unique identifier makes sense, as the tab ID remains constant within the same tab...

Using a path parameter instead of a search parameter won't solve the issue you stated. They're functionally equivalent and using one over the other is just a matter of preference. In this specific case, I prefer a search param.

sebastien-comeau:
> Using a path parameter instead of a search parameter won't solve the issue you stated. They're functionally equivalent and using one over the other is just a matter of preference. In this specific case, I prefer a search param.

I asked this to LLM and I agree:

> I always treat search parameters as an optional part of the URL. Is that the right approach?
> 

> > Yeah, you're totally fine treating search params as optional. By definition, search parameters (the `?key=value` part of a URL) are meant to be extra, optional data—they don’t affect the core path structure of a URL.
> > 
> > However, a couple of things to keep in mind:  
> > 1. **Don't rely on them always being there** – Your app should work even if a search param is missing. Default values can help.  
> > 2. **Use them for transient state** – They're great for filtering, sorting, pagination, and non-critical UI state but shouldn't be the sole source for required data.  
> > 3. **Be cautious with encoding** – Some values might need to be properly encoded/decoded (e.g., spaces become `%20` or `+` in URLs).  
> > 4. **Avoid sensitive data** – Since search params are visible in the URL, don’t put anything sensitive in them.
> > 
> > Are you dealing with a case where you're unsure if a search param should be required?
>

gregory-j-baker:
> I always treat search parameters as an optional part of the URL. Is that the right approach?

Exactly. The tab id *is* optional. You must be able to hit the URL without it and still have the application function as expected... not return a 404 response (which would happen if you used a path param).

sebastien-comeau (approved):

PR #306, opened 2025-02-28
fix(frontend): fix incorrect error code

gregory-j-baker:
## Summary

The `getLocalizedLanguageOfCorrespondence()` functions were using the wrong error codes when throwing.

## Types of changes

- [x] 🐛 **bugfix** -- non-breaking change that fixes an issue
- [ ] ✨ **new feature** -- non-breaking change that adds functionality


sebastien-comeau (approved):
Good catch. You can also remove `NO_OPTION_SET_FOUND` that is not used anymore since my latest PR #307

gregory-j-baker:
> Good catch. You can also remove `NO_OPTION_SET_FOUND` that is not used anymore since my latest PR #307

Sure.. $25.

PR #366, opened 2025-03-07
refactor(frontend): Refactor validation errors

Dario-Au:
## Summary

[AB18963](https://dev.azure.com/ESDCCM/FSIR/_workitems/edit/18963) Refactor validation errors
Validation refactored to return error keys rather than pre-translated strings

## Types of changes

## Checklist

### DIFF
--- a/frontend/app/routes/protected/person-case/birth-details.tsx
+++ b/frontend/app/routes/protected/person-case/birth-details.tsx
@@ -26,6 +26,7 @@ import { getTranslation } from '~/i18n-config.server';
 import { handle as parentHandle } from '~/routes/protected/person-case/layout';
 import { getStateRoute, loadMachineActor } from '~/routes/protected/person-case/state-machine';
 import type { BirthDetailsData } from '~/routes/protected/person-case/types';
+import { getSingleKey } from '~/utils/i18n-utils';
 import { REGEX_PATTERNS } from '~/utils/regex-utils';
 import { trimToUndefined } from '~/utils/string-utils';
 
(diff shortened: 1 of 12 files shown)


frontend/app/utils/string-utils.ts
```diff
@@ -38,6 +40,15 @@ export function trimToUndefined(str: string | undefined): string | undefined {
...
 
+/**
+ * Retrieves the key from a fetcher error.
+ * @param err - The error from which to extract the key.
+ * @returns A ResourceKey, or undefined if no key is found.
+ */
+export function getErrorKey(err: [string, ...string[]] | undefined): ResourceKey | undefined {
+  return err?.at(0);
+}
+
```
gregory-j-baker, line 51: I have a couple of comments for this function..

1. I think this would feel more at home in `i18n-utils`
2. **Nitpick:** `err` should be plural, since it's an array

gregory-j-baker: Actually.. a few more comments..

If `err.length > 1`, what happens to the rest of the data? It's just ignored?

I wonder if this would be better as this: (not sure about the syntax, but you should get the idea)

```ts
function asResourceKey(str: string | undefined): ResourceKey | undefined {
  return str as ResourceKey;
}
```

and then called using `asResourceKey(errors?.at(0))`..

the reason is, it would provide flexibility to the caller just in case `errors.at(1)` is important.. (ie: it let's the caller discard the additional elements, rather than doing it here).

PR #373, opened 2025-03-10
feat(frontend): create SIN application request

faiza-jahanzeb:
## Summary

[AB18736](https://dev.azure.com/ESDCCM/FSIR/_workitems/edit/18736)
[AB19029](https://dev.azure.com/ESDCCM/FSIR/_workitems/edit/19029)
Create SIN application request to call Create Case API Interop proxy.

## Types of changes

## Checklist

<details>

  ``` shell
  npm run lint:check
  npm run format:check
  ```
</details>

<details>

  ``` shell
  npm run test
  npm run test:e2e
  ```
</details>

## Additional Notes

If this PR introduces significant changes, explain your reasoning and provide any necessary context here. Feel free to include diagrams, screenshots, or alternative approaches you considered.

## Screenshots (if applicable)

Provide screenshots or screen-recordings to help reviewers understand the visual impact of your changes, if relevant.


frontend/app/.server/shared/services/sin-application-service.ts
```diff
@@ -0,0 +1,337 @@
+import { serverEnvironment } from '~/.server/environment';
+import type { SinApplicationRequest } from '~/.server/shared/api/interop';
+import type { InPersonSinApplication } from '~/routes/protected/person-case/types';
+
+export function mapInPersonSINCaseToSinApplicationRequest(
```
gregory-j-baker, line 5: This service directly utilizes the `InPersonSinApplication` type, however, `InPersonSinApplication` was intentionally scoped to the routes/protected/person-case/ module. This scoping is important to ensure proper separation of concerns as data is collected, then mapped, then sent to the API.

To maintain this separation of concerns, I recommend the following approach:

1. this function should accept a defined data structure as its input that is scoped to this service module.
1. within the person-case route, when submitting, transform the data from the `InPersonSinApplication` type into the format expected by this service.
1. if the data structures used by the service closely resemble the types in `routes/protected/person-case/types.d.ts`, you can move those types into the service module and then import them into the person-case module.

The reason we do it like this is because the data required to complete the form might not match the data required to call the API. Sometimes it will, but not always.

And the reason the data should be converted from **form data** to **service data** in the route module is because the route module knows how to call the service, but the service should never know anything about who or what is calling it (ie: it doesn't know about the route module).

frontend/app/routes/protected/person-case/review.tsx
```diff
@@ -60,6 +66,14 @@ export async function action({ context, params, request }: Route.ActionArgs) {
     }
 
     case 'next': {
+      try {
+        const response = await sinApplicationService.submitSinApplication(sinApplicationRequest);
+        //TODO: store the response in session
+        console.log('SIN Application submitted successfully:', response);
+      } catch (err) {
+        console.log('Error submitting SIN application:', err);
+      }
```
gregory-j-baker, line 75: This code should be moved to the state machine's `entry` action for its final state. For now, you can continue to keep it here and we can look at moving it once the API we have to call stabilizes.

frontend/app/.server/shared/services/sin-application-service.ts
faiza-jahanzeb: not sure what do you mean. 
There is already a data structure for the service [here](https://github.com/DTS-STN/future-sir/blob/main/frontend/app/.server/shared/api/interop/types.gen.ts)
This file is just a mapper to map InPersonSinApplication to SinapplicationData (not an actual service)
While the service is [here](https://github.com/DTS-STN/future-sir/blob/main/frontend/app/.server/domain/person-case/services/sin-application-service.ts)
What am I missing? Are you saying to move this function mapInPersonSINCaseToSinApplicationRequest  in the route module?

gregory-j-baker: We typically structure applications in layers:

image

We do this to help facilitate looser coupling within the application. In the diagram above, the UI layer knows that the service layer exists, but the service layer does not know that the UI layer exists.. it only knows about the data layer.

(we don't have a data layer, but we *could*)

So your `sin-application-service.ts` module should never import anything from `~/routes/`. When you call the service, the route should convert the `InPersonSinApplication` data into a `SinApplicationRequest` object first, and then call the service.

One way to know if the code is adequately decoupled is to ask: "*if I delete this route, do I have to modify code elsewhere in the application?*" If the answer is yes, then the code is probably tightly coupled to something it shouldn't be.

In your code, if we delete the in-person route, then we would have to rewrite the service module. We want to avoid this sort of thing whenever possible.

faiza-jahanzeb: Thanks for explaining in detail
updated my pr, I hope this is what you mean... (I moved InPersonSinApplication in types.d.ts to the server module and then import them into the person-case module)
I wasn't quiet sure about that, because normally on server side, we have dtos (for defining data structure), dto mappers (to map incoming data to the data structure utilized by the service), and services that are separate from the UI layer.

gregory-j-baker: What you did is good except:

1. I think the types would be better off in the `sin-application-service.ts` file. A `.d.ts` file was used in the route because there we were doing [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation), but that doesn't apply here.
2. The types should be reviewed to make sure they make sense. The original types were created to hold user input for a single form, by moving them to the service, they have a different responsibility. For example, it might not make sense to have 10+ types anymore.. maybe we only need 1 or a couple/few.
  
I think we can merge the code for now (once you move the types into the service), but I do think we should review the data structure and adjust if necessary in a later PR.

gregory-j-baker (approved):

